PR #12, opened 2024-03-26
Google gemini

vitaglianog:
The purpose of this PR is to add support for image processing with more LLM services - specifically including the Google Gemini vision model.

### Summary of changes to extend image processing: 
 - Added a new Prompt Strategy IMAGE_TO_TEXT, which is used when the plan involves a File->ImageFile conversion.
 - Include the Google Gemini model card, and enable its use through DSPY.
 - Generalized the image processing module to also include Google Gemini.
 - Add a new test module to test the image processing scenario.
 - Warning: updated requirements to update dspy for the latest version. This package is not yet available on PyPi, so it must be installed manually. This is required due to a bug in previous versions involving google-gemini. 

### Detailed changes per file:

 - .gitignore
  - Added .env file that contains secret API keys 
  - Added folders build/* and vscode/* for development.
  
 - pyproject.toml
   - Update Dspy version due to bugfix related to Google-Gemini
   - Included google-generativeai package
   - Fix typo
  
 - src/cli/cli_main.py
   - Added the 'google' choice for LLM service
  
 - src/palimpzest/constants.py
   - Added the two model cards for Gemini and Gemini Vision
   - Added a new Prompt Strategy IMAGE_TO_TEXT
  
 - src/palimpzest/operators/operators.py
   - Extend models to include Google Gemini if API key is found in environment
   - Added an assertion that checks for model availability when creating physical plans. I suppose it could be made less strict but I was running into errors if no API were set.

 - src/palimpzest/operators/physical.py
   - If the plan involves a File->ImageFile conversion, set the prompting strategy to IMAGE_TO_TEXT. Rather than always selecting the GPT_4 model, check whether the specified model is from OpenAI or Google - and then pick accordingly. Maybe we should still go more general than this ?

 - src/palimpzest/operators/solver.py
   - Moved code the openai specific function to a general "describe_image" function
   - Moved File->ImageFile conversion from hardcoded function to the _makeLLMTypeConversionFn
   - If the prompt strategy is IMAGE_TO_TEXT, encode the image as a b64 and pass it to the describe_image function along with the chosen model.

 - src/palimpzest/tools/dspysearch.py
   - If the chosen model is from Google, check for the GOOGLE_API_KEY, and set the dspy model accordingly.

 - src/palimpzest/tools/generic_image_converter.py
   - This module generalizes the previous openai specific module. It is not much more refined than the previous one, but it includes a switch to check if the model is GPT or Google Gemini. I would refactor it to make it more general, but the api/modules from the two services are not very comparable.
 
 - tests/context.py
   - This convenience module is used to import a ".env" file with the environment variables for the tests. In testing modules, simply import this module to have access to the environment variables.

 - tests/imageDemo.py
   - Abridged version of simpleDemo.py to quickly test for the image processing scenario.
   - Added a --no-cache argument that removes the content of the cache dir before running the test.


mdr223 (approved):
Great work Gerardo! This looks good to me.

PR #13, opened 2024-03-26
Refactor profiling so that we can have multiple plans at once, some o…

mikecafarella:
…f which might profile and others which don't.  Also add preliminary framework for query reoptimization

### DIFF
--- a/src/palimpzest/operators/physical.py
+++ b/src/palimpzest/operators/physical.py
@@ -25,17 +25,16 @@ class PhysicalOp:
     synthesizedFns = {}
     solver = Solver(verbose=LOG_LLM_OUTPUT)
 
-    def __init__(self, outputSchema: Schema) -> None:
+    def __init__(self, outputSchema: Schema, shouldProfile = False) -> None:
         self.outputSchema = outputSchema
         self.datadir = DataDirectory()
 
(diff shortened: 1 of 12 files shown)


src/palimpzest/operators/physical.py
```diff
@@ -632,7 +631,8 @@ def iteratorFn():
 
                 # if we're profiling, then we still need to yield candidate for the profiler to compute its stats;
                 # the profiler will check the resultRecord._passed_filter field to see if it needs to be dropped
-                elif Profiler.profiling_on():
+                #elif Profiler.profiling_on():
+                else:
```
mdr223, line 635: are we switching `elif` -> `else` b/c now we're always profiling?

src/palimpzest/operators/physical.py
```diff
@@ -778,7 +778,8 @@ def iteratorFn():
 
                         # if we're profiling, then we still need to yield candidate for the profiler to compute its stats;
                         # the profiler will check the resultRecord._passed_filter field to see if it needs to be dropped
-                        elif Profiler.profiling_on():
+                        #elif Profiler.profiling_on():
+                        else:
```
mdr223, line 782: ditto to question above in FilterOp

mdr223 (approved):
LGTM

PR #14, opened 2024-03-28
Cardinality and funcs

mikecafarella:


src/palimpzest/sets/sets.py
```diff
@@ -213,9 +234,15 @@ def filterByStr(self, filterCondition: str, desc: str="Apply filter(s)") -> Data
 
         return self.filter(f, desc)
 
-    def convert(self, newSchema: Schema, desc: str="Convert to new schema") -> Dataset:
+    def convert(self, newSchema: Schema, cardinality: str = None, desc: str="Convert to new schema") -> Dataset:
```
mdr223, line 237: See note I will leave in `simpleDemo.py`

tests/simpleDemo.py
```diff
@@ -45,6 +47,49 @@ def buildMITBatteryPaperPlan(datasetId):
...
+    pdfLink = pz.Field(desc="The link to the PDF of the paper", required=True)
+
+def downloadVLDBPapers(vldbListingPageURLsId, outputDir):
+    """ This function downloads a bunch of VLDB papers from an online listing and saves them to disk.  It also saves a CSV file of the paper listings."""
+    policy = pz.MaxQuality()
+
+    # 1. Grab the input VLDB listing page(s) and scrape them for paper metadata
+    tfs = pz.Dataset(vldbListingPageURLsId, schema=pz.TextFile, desc="A file full of URLs of VLDB journal pages")
+    urls = tfs.convert(pz.URL, desc="The actual URLs of the VLDB pages", cardinality="oneToMany")  # oneToMany=True would be nice here.   
+    htmlContent = urls.map(pz.DownloadHTMLFunction())
```
mdr223, line 63: A minor note, which I think dovetails w/your comment in `corelib/functions.py`, is that we have now defined a `.convert()` and a `.map()` function for `pz.Dataset`. I feel like we should combine these functions into one, and if the user passes in a UDF then we execute it, and if the user does not pass in a UDF then we run the usual `.convert()` function.

mdr223: If you agree with this^ idea, a secondary issue would be to decide between keeping `.map()` or `.convert()` for a naming purpose. I have some thoughts on the pros/cons of each, but I'll wait to see if you agree first with the idea that we should merge the two.

tests/simpleDemo.py
```diff
@@ -45,6 +47,49 @@ def buildMITBatteryPaperPlan(datasetId):
...
+    vldbPaperListings = htmlContent.convert(VLDBPaperListing, desc="The actual listings for each VLDB paper", cardinality="oneToMany")
```
mdr223, line 64: While I don't think we should change it now, I think asking the user to provide a `cardinality` is a bit odd (it seems like something we should handle for them, but I understand that we can't know the `cardinality` a priori).

One half-baked idea for an alternative may be to introduce a `.flatten()` syntax, and if we see, e.g.:
```python
ds = pz.Dataset(...)
ds_new = ds.convert(...)
ds_newer = ds_new.flatten()
```
Then behind the scenes we effectively convert `ds.convert(...).flatten()` into `ds.convert(..., cardinality="oneToMany")`.

mdr223 (approved):
Left some food-for-thought comments; approved.

tests/simpleDemo.py
mikecafarella: I mainly think that managing cardinality is a core part of the data model of the universe. Sometimes I am OK leaving it up to the system, but in many cases I have strong constraints I want the system to enforce.  (Like: exactly one title, and >=1 authors)

Though I do think it's a bit weird to have it in the query syntax, rather than the schema declarations, where we ask the user to put most of their data model decisions.

mikecafarella: I don't think the current method is perfect and am open to suggestions

PR #15, opened 2024-03-28
Refactor as Prelude to Coming PR for Stats Processing + Use in Execution and Cost Est.

mdr223:
As I part of my work on (A) fixing the profiling stats collection (B) creating a class for processing profiling stats and (C) making use of these stats in cost estimation, I did a lot of refactoring to help standardize (and simplify) certain functions.

To help isolate the changes related to A, B, and C, I've decided to split this work into two PRs:
- The first one (this PR) contains just the refactoring of code
- The second one (which I think will follow tonight or tomorrow at the latest) will contain changes A, B, and C outlined above

This PR contains the following changes, in rough order of importance:

1. Creation of a `QueryStrategy` enum
    - This enum presents a new degree of freedom for our optimizer to consider
    - I noticed that the `Solver` code around LLM induce tasks was getting hairy, and I also realized that the choice of what "query strategy" we use (e.g., conventional, bonded, code gen, bonded w/fallback to conventional, etc.) would change our estimates of the time, cost, and potentially accuracy of these LLM induce tasks
    - For now, I just default to using `QueryStrategy.BONDED_WITH_FALLBACK` everywhere, but in a future PR this will definitely be an optimization dimension that I consider

2. The creation of a `BaseGenerator` class, which is sub-classed to produce `Generator`s
    - A `Generator` exposes a `generate()` function, which is responsible for passing a `prompt` and some `input` to an LLM and returning the output
    - Implemented in `src/palimpzest/generators`, this maps roughly 1-to-1 with code we previously had scattered across `dspysearch.py`, `dspyadaptors.py`, and `generic_image_converter.py`
    - The `BaseGenerator` standardizes the expected output of any `Generator` to include a stats dictionary
    - This will hopefully make it a bit easier/cleaner to write generation code which supports multiple prompt strategies at once

4. Creation of a `TaskDescriptor` dataclass
    - In the short-term, the benefit is mostly to make the passing of tuples and input arguments in `solver.py` a bit more readable
    - In the long-term, we can subclass `TaskDescriptor` to have e.g. `InduceTaskDescriptor`, `FilterTaskDescriptor`, `RAGTaskDescriptor`, etc.; this will allow us to formalize the fields that each task needs and it can make it easier for us to code against this formalism

I'm first going to merge Mike's [current PR](https://github.com/mikecafarella/palimpzest/pull/14) into `main`. Then I'll merge `main` into this branch and merge back.

Once that's done, I'll finish my partially completed work on A, B, and C.


src/palimpzest/generators/generators.py
```diff
@@ -0,0 +1,234 @@
...
+
+        # collect statistics on prompt, usage, and timing
+        stats = {
+            'api_call_duration': end_time - start_time,
+            'prompt': prompt,
+            'usage': usage,
+            'finish_reason': finish_reason,
+        }
+
+        return answer, stats
```
mikecafarella, line 234: I think this is good. How would we implement the ASKEM equation extractor using this framework? @Tranway1

src/palimpzest/solver/query_strategies.py
```diff
@@ -0,0 +1,258 @@
...
+# from palimpzest.profiler import Stats
+from palimpzest.solver.task_descriptors import TaskDescriptor
+
+from typing import Any, Dict, List, Tuple
+
+import base64
+import json
+
+
+def _construct_query_prompt(td: TaskDescriptor, doc_type: str, generate_field_names: List[str]) -> str:
```
mikecafarella, line 13: This is better than what we used to have, but seems like we still need an intermediate representation of data operations so we can combine novel optimizations in a more principled way than we do now.

This is worth spending some time brainstorming and discussing.

src/palimpzest/solver/solver.py
```diff
@@ -71,273 +86,180 @@ def _fileToPDF(candidate: DataRecord):
...
+                if shouldProfile:
+                    stats["bondedQuery"] = bonded_query_stats
+                    for dr in drs:
+                        dr._stats[td.op_id] = stats
+
+                return drs
+
+            # TODO
+            elif td.query_strategy == QueryStrategy.CODE_GEN:
+                raise Exception("not implemented yet")
```
mikecafarella, line 184: Should this be flagged for Zui as a place to add his code?

src/palimpzest/solver/query_strategies.py
```diff
@@ -0,0 +1,258 @@
...
+            query_stats["out_fields"] = generate_field_names
+
+        # TODO
+        elif td.prompt_strategy == PromptStrategy.ZERO_SHOT:
+            raise Exception("not implemented yet")
+
+        # TODO
+        elif td.prompt_strategy == PromptStrategy.FEW_SHOT:
+            raise Exception("not implemented yet")
+
```
mikecafarella, line 176: Where would Chunwei add his optimization that figures out which parts of the input are most important for a particular output?

src/palimpzest/operators/physical.py
```diff
@@ -4,9 +4,10 @@
 from palimpzest.corelib.schemas import ImageFile
 from palimpzest.datamanager import DataDirectory
 from palimpzest.elements import *
-from palimpzest.elements import Any
-from palimpzest.solver import Solver
-from palimpzest.tools.profiler import Profiler
+from palimpzest.elements import Any # TODO: can we delete?
```
mikecafarella, line 7: In terms of overall code cleanliness, this is my least favorite file in the project.  Let's consider how we might redo the class hierarchy and reduce the amount of code needed for each new physical operator.

mikecafarella (commented):
Overall there's a lot of changes but I think they are positive steps. In terms of correctness, LGTM, but of course let's make sure it works on all the simpleDemo test cases before committing.

Let's flag a few areas as targets for Chunwei, Zui, and others to insert optimizations.

src/palimpzest/elements/records.py
```diff
@@ -1,33 +1,23 @@
...
-        #if self._shouldProfile:
-            # self._stats is a dictionary mapping a physical opId --> a stats dictionary
-            # for the LLM operation(s) performed during that physical operation (if applicable);
-            # the self._stats dictionary is updated in solver.py as the physical operation is performed;
-            # some timing statistics will also be computed inside of profiler.py when the DataRecord
-            # is intercepted before being passed on to the next operation.
-        self._stats = {}
-        self._state = {}
+        # unique identifier for the record
+        self.uuid = uuid.uuid4()[:MAX_UUID_CHARS]
```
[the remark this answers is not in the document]
mdr223: I didn't think about that, but that is a good point! Right now one kind of clumsy thing about this setup is that when we create a new DataRecord for an induce operation, we often do something like:
```python
fn(candidate: DataRecord):
    dr = DataRecord(outputSchema)
    # do induce and update dr fields
```
So I've need to manually find all the places in the code where this happens and change them to:
```python
fn(candidate: DataRecord):
    dr = DataRecord(outputSchema)
    dr.parent_uuid = candidate.uuid
    # do induce and update dr fields
```
I think this is possibly a good candidate (pun not intended) for creating an intialization method like `.from_record()` which would automatically set the `parent_uuid` and copy any duplicate fields:
```python
fn(candidate: DataRecord):
    dr = DataRecord.from_record(candidate, schema=outputSchema)
    #^ this sets parent_uuid automatically and copies duplicate fields
    # do induce and update non-duplicate dr fields
```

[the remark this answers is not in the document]
mdr223: (or maybe `.from_parent()` instead of `.from_record()`, idk)

src/palimpzest/generators/dspy_utils.py
```diff
@@ -1,43 +1,88 @@
...
+def gen_signature_class(instruction, context_desc, question_desc, answer_desc):
+    class QuestionOverDoc(dspy.Signature):
+        __doc__ = instruction
+        context = dspy.InputField(desc= context_desc)
+        question = dspy.InputField(desc= question_desc)
+        answer = dspy.OutputField(desc= answer_desc)
+    return QuestionOverDoc
+
+def gen_filter_signature_class(doc_schema, doc_type):
+    instruction = f"Answer condition questions about a {doc_schema}."
```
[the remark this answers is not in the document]
mdr223: Very much agree, I'd be interested in looking into existing packages and seeing if any would slot in nicely here. I assume someone has already invented this wheel so we should be able to borrow from them. (Even though I don't love LangChain, their `PromptTemplate` class is not terrible, maybe that's worth a look).

src/palimpzest/generators/dspy_utils.py
```diff
@@ -1,43 +1,88 @@
...
         if any(keyword in self.model.lower() for keyword in ["inst", "instruct"]):
             self.use_inst_template = True
 
         stop_default = "\n\n---"
 
-#        print("Stop procedure", stop_default)
+        # print("Stop procedure", stop_default)
         self.kwargs = {
             "temperature": 0.0,
             "max_tokens": 8192,
```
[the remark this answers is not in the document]
mdr223: Agreed 👍

src/palimpzest/solver/query_strategies.py
mdr223: Yeah, I don't think this is finished work, I was having a hard time deciding what the interface between `Solver`, `Generator`, and the functions in this file should be. I think this is worthy of a small meeting early next week to brainstorm.

src/palimpzest/solver/solver.py
mdr223: Yes, in PR2 I think I've added such a note.

src/palimpzest/solver/query_strategies.py
mdr223: Good question, and it may be something I can answer better after finishing PR2, since some of that will depend on how the re-optimization code works.

src/palimpzest/operators/physical.py
mdr223: 100% agreed, the simplest thing I can think of is to have the parallel induce and filter operations subclass their non-parallel counterparts, but even that only brings things from ~1000 lines --> ~900. There's definitely more we could to to clean this up.

src/palimpzest/elements/records.py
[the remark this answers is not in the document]
mikecafarella: Agreed, this is a good idea

src/palimpzest/generators/dspy_utils.py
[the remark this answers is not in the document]
mikecafarella: I think LangChain is the right choice. It's simple enough (at least it was, the last time I looked) that we can have confidence about its rewritings. It's hard for me to know what DSPy is doing.

src/palimpzest/generators/dspy_utils.py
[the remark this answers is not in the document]
mikecafarella: I think the existence of all this stuff suggests there is a useful abstraction prior to prompt generation that we haven't invented yet.

PR #16, opened 2024-03-29
Enhance and Utilize Profiling Data for Query Re-Optimization

mdr223:
This PR enables the `Execution` class to:
1. Extract profiling data from the execution of a query plan (ideally with a `limit`)
2. Compute improved cost estimates for each physical operator based on the sample of records executed in the limit query
3. Choose which (potentially new) query plan to run based on the improved cost estimation

I have tested this work on the `enronoptimize` task, which can be executed as follows:
```bash
$ python tests/simpleDemo.py --datasetid enron-tiny --task enronoptimize --policy mincost --profile --verbose
```

I have also implemented some skeleton code where Zui (after merging in this branch) could plug-in his code generation work. I annotated the two main entrypoints in the code for Zui with the string: `# TODO: @Zui`. @Magolor 

My current status/progress on this PR is roughly the following:
- ✅ Standardize `Solver` to annotate records with `Stats` objects everywhere
- ✅  Collect stats objects (and record states) into better data structure in `profiler.py`
- ✅  Implement `StatsProcessor` class with functions that automatically compute views of interest
- ✅  Use profile data + `StatsProcessor`  inside of `Execution` (and specifically inside of `logicalTree.createPhysicalPlanCandidates()` to re-optimize plans

I believe this PR is ready for review and initial feedback.

My own primary critique at the moment is that the re-optimization will be seriously flawed when the model/plan used to generate the initial execution samples is (significantly) different from a different model/plan which we consider during re-optimization. At a minimum, I believe that one more piece of work I should add to this PR is either:
1. Supporting the execution of multiple plans to generate initial samples for cost estimation with different models
2. Supporting some form of translation from cost estimation samples generated by model A to samples generated by model B

For an example of the latter of these^ two ideas, suppose I use GPT-4 in my sample plan in the `Execution` class. Then, if I want to consider a plan which uses GPT-3.5 during re-optimization, I could:
- Multiply the `usd_per_record` by the (dollar) cost ratio between processing tokens with GPT-4 and GPT-3.5
- Multiply the `est_num_output_tokens` by some constant empirical ratio of the models' relative verbosity (e.g. if GPT-3.5 on average outputs more text than GPT-4, this ratio could be 1.25 instead of 1.0)

Multiply other fields (e.g., time_per_record) by similar ratios


src/palimpzest/corelib/functions.py
```diff
@@ -38,6 +38,7 @@ def html_to_text_with_links(self, html):
     def map(self, dr: DataRecord):
         textcontent = requests.get(dr.url).text
         dr2 = DataRecord(self.outputSchema)
```
mikecafarella, line 40: Seems kind of error-prone to have a DR constructor that requires we always follow it up with a parent assignment.  Shouldn't parent assignment be required by default?

src/palimpzest/operators/operators.py
```diff
@@ -167,38 +165,29 @@ def _createPhysicalPlans(self, max: int=10, shouldProfile: bool=False) -> Physic
...
             for physicalPlan in logicalPlan._createPhysicalPlans(shouldProfile=shouldProfile)
         ]
-        t_physical = time.time()
-        # print(f"Time to create physical plans: {t_physical - t_logical:.2f}")
 
         # estimate the cost (in terms of USD, latency, throughput, etc.) for each plan
         plans = []
         for physicalPlan in physicalPlans:
-            planCost = physicalPlan.estimateCost()
+            planCost = physicalPlan.estimateCost(cost_estimate_sample_data=cost_estimate_sample_data)
```
mikecafarella, line 184: If the dictionary is not provided, is the intention that estimateCost() use historical data?

src/palimpzest/operators/operators.py
```diff
@@ -167,38 +165,29 @@ def _createPhysicalPlans(self, max: int=10, shouldProfile: bool=False) -> Physic
 
         return physicalPlans
 
-    def createPhysicalPlanCandidates(self, sampler: Sampler=None, shouldProfile: bool=False) -> List[PhysicalPlan]:
+    def createPhysicalPlanCandidates(self, cost_estimate_sample_data: List[Dict[str, Any]]=None, shouldProfile: bool=False) -> List[PhysicalPlan]:
```
mikecafarella, line 169: It's a drag to have this stuff in just a Dict....  should this be a proper class, or is the flexibility of this needed?

src/palimpzest/operators/physical.py
```diff
@@ -27,24 +26,46 @@ class PhysicalOp:
...
+        # simply return stats for this operator if there is no source
+        if self.shouldProfile and self.source is None:
+            return self.profiler.get_data()
+
+        # otherwise, fetch the source operator's stats first, and then return
+        # the current operator's stats w/a copy of its sources' stats
+        elif self.shouldProfile:
+            source_operator_stats = self.source.getProfilingData()
+            operator_stats = self.profiler.get_data()
+            operator_stats.source_op_stats = source_operator_stats
```
mikecafarella, line 61: Again, should we enforce this by requiring that source_operator_stats be passed into profiler.get_data()?

mikecafarella (commented):
Overall I think this is reasonable, although the size is so daunting I am a bit concerned I missed something.

However whatever flaws I've missed, I think this is the way forward from what we have today.

src/palimpzest/operators/operators.py
mdr223: If the list[dict] is not provided then it should use our estimates w/zero prior information (i.e., just using constants for things like selectivity, read time, etc.)

src/palimpzest/corelib/functions.py
mdr223: Good catch, I added a `parent_uuid` kwarg and didn't update it here, will push that in a second.

src/palimpzest/operators/physical.py
mdr223: Longer term yes, although I think this will complicate a few things to change it now, so let's mark this as a tech debt to resolve post-preprint

PR #17, opened 2024-04-02
Add a new group by operator

srmadden:
Add code to support group bys.  This should supersede the existing aggregation operator, but I have not removed that code in this pull request.

I have added several examples of aggregation to simpleDemo.py.  For example:

```
def enronGbyPlan(datasetId):
    emails = pz.Dataset(datasetId, schema=Email)
    ops = ["count"]
    fields = ["sender"]
    groupbyfields = ["sender"]
    gbyDesc = GroupBySig(groupbyfields, ops, fields)
    groupedEmails = emails.groupby(gbyDesc)
    return groupedEmails
```

Here we count emails by sender.  The GroupBySig class is a class that encapsulates the signature of the group by and contains utility functions for getting the output schema of the group by operation.

Note passing in an empty (not null) array for the groupbyfields variable here would result in performing a ungrouped aggregate.

Currently only count and average are supported but adding new aggregates should be straightforward.  A useful extension might be to add an LLM-based aggregate where the group of records is sent to an LLM to perform the aggregation over them.

### DIFF
--- a/src/palimpzest/datamanager/datamanager.py
+++ b/src/palimpzest/datamanager/datamanager.py
@@ -217,7 +217,6 @@ class DataDirectory(metaclass=DataDirectorySingletonMeta):
         """Return a cached result."""
         if not cacheId in self._cache:
             return None
-
         cachedResult = pickle.load(open(self._cache[cacheId], "rb"))
         def iterateOverCachedResult():
             for x in cachedResult:
@@ -251,7 +250,11 @@ class DataDirectory(metaclass=DataDirectorySingletonMeta):
     def closeCache(self, cacheId):
         """Close the cache."""
         filename = self._dir + "/data/cache/" + cacheId + ".cached"
-        pickle.dump(self._tempCache[cacheId], open(filename, "wb"))
+        try:
+            pickle.dump(self._tempCache[cacheId], open(filename, "wb"))
+        except pickle.PicklingError:
+            print("Warning: Failed to save cache due to pickling error")
+            os.remove(filename)
         del self._tempCache[cacheId]
         self._cache[cacheId] = filename
 
--- a/src/palimpzest/elements/groupbysig.py
+++ b/src/palimpzest/elements/groupbysig.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 from typing import Any, Dict
-from palimpzest.elements import Field, OperatorDerivedSchema
+from palimpzest.elements import Field, OperatorDerivedSchema, Schema
 
 #signature for a group by aggregate that applies
 # group and aggregation to an input tuple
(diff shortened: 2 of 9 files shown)


src/palimpzest/elements/groupbysig.py
```diff
@@ -0,0 +1,50 @@
+from __future__ import annotations
+from typing import Any, Dict
+from palimpzest.elements import Field, OperatorDerivedSchema
+
+#signature for a group by aggregate that applies
+# group and aggregation to an input tuple
+class GroupBySig:
```
mikecafarella, line 7: Should GroupBySig take as input the Schema that it will be applied to?  If it did, then we could test for syntactic compatibility at the moment the GroupBySig is defined.  E.g., if you try to group by a field that doesn't exist in the target schema, then you can't create the GroupBySig.

Is it useful to have GroupBySigs that exist independently of the Schema that is being grouped-by?  I am not sure about that.

I am honestly not sure what's best here.

src/palimpzest/operators/physical.py
```diff
@@ -825,6 +825,116 @@ def _passesFilter(self, candidate):
...
+                        group = group + (getattr(r,f),)
+                    if group in aggState:
+                        state = aggState[group]
+                    else:
+                        state = []
+                        for fun in self.gbySig.aggFuncs:
+                            state.append(agg_init(fun))
+                    for i in range(0,len(self.gbySig.aggFuncs)):
+                        fun = self.gbySig.aggFuncs[i]
+                        field = getattr(r, self.gbySig.aggFields[i])
```
mikecafarella, line 913: If the GroupBySig is not correctly defined, then I think this is the line on which we would expect to see a problem.  The getAttr() will fail here.  Maybe we can have more explicit field compatibility checking so the user gets a more informative error message?

src/palimpzest/datamanager/datamanager.py
```diff
@@ -251,7 +250,11 @@ def appendCache(self, cacheId, data):
     def closeCache(self, cacheId):
         """Close the cache."""
         filename = self._dir + "/data/cache/" + cacheId + ".cached"
-        pickle.dump(self._tempCache[cacheId], open(filename, "wb"))
+        try:
+            pickle.dump(self._tempCache[cacheId], open(filename, "wb"))
```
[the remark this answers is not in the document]
mikecafarella: Basically pickle was a fast early hack.  But if we want caching to work across users someday, we need to come up with a more well-engineered serialization and caching scheme

mikecafarella (commented):
Overall LGTM and I like the design, but have a few questions around (1) whether GroupBySigs are defined in terms of an existing Schema or should exist independently, and (2) error checking

src/palimpzest/elements/groupbysig.py
srmadden: Yes this is probably useful to check here -- I was sort of in the mindset of a query planner where this kind of semantic compatibility would be handled by the layer above but since this is the interface programmers directly write having some decent error checking is a good idea.   I can add that.

src/palimpzest/operators/physical.py
srmadden: Yes agreed, some error checking here would be good.

mikecafarella (approved):
LGTM!

PR #18, opened 2024-04-12
Support for Tables + biofabric demo

vitaglianog:
### Summary of changes
This PR subsumes a few pieces of work:
 - Includes new printing in order for physical plans
 - Includes new schema for Excel files and Tables, along with hardcoded functions to convert from File to XLS and from XLS to Table (one to many potentially)
 - Includes code to run the biodata fabric demo with table filtering
 - Fixed some bugs related to the use of log probability profiling
 - Fixed some bugs related to google-gemini api

### Per-file changes:

- pyproject.toml
   - Removed comments about dspy for google-gemini 
   - Added openpyxl dependency

- src/palimpzest/operators/operators.py
  Removed code lines that switch from Google Gemini to Openai since dspy 2.4.1 is now pushed to pypi.
  
- src/palimpzest/operators/physical.py
  Removed code lines that switch from Google Gemini to Openai since dspy 2.4.1 is now pushed to pypi.

- src/palimpzest/execution
  - __init__.py
    - Added SimpleExecution

  - execution.py
    - Small bugfix: if profiling was never done before and verbose==True, there would be a missing folder.
    - Added new SimpleExecution class that does not pre-sample data. This is useful for debugging and testing.
    - Added a new convenience function to un-nest tuples in the output of the execution.
    - Added a new graphicDump method to have a visual representation of schema changes.

- src/palimpzest/elements
  - __init__.py
    - Added pzlist to init file

 - records.py
    - Added "..." to the str representation 

 - pzlist.py
    - Rewrote import to avoid circular dependency

- src/palimpzest/generators/generators.py
  - Debugging made easier by adding reraise=True to the @retry annotation, now the internal error message is shown.
  - Fixed signature mismatch for Google in DsPY.
  - Fixed a bug because Google API does not provide log probabilities.
  - Commented while function to avoid infinite loop when matching answer with tokens
  

- src/palimpzest/solver
  - solver.py
    - Added hard-coded function to convert from Download object to File object
    - Added hard-coded function to convert from XLS file to CSV file object

- src/palimpzest/corelib
  - schemas.py
    - Added new type of File: XLSFile 
    - Added TableRow and Table 

  - functions.py
    - WIP to allow redirects and javascript parsing
    
### Future work 
 - The PDF are usually too large, so there is a fixed limit 
 - Currently the page download is hardcoded, to retrieve the URLs of the tabular files
 - The conversion from File to XLS is hardcoded (reads sheet names)
 - The conversion from XLS to Table is hardcoded
 - The filter of the tables cannot take the full table due to context length limitations, for this reason when we convert to a table we only take the first 10 rows of the table.


src/palimpzest/elements/pzlist.py
```diff
@@ -1,4 +1,4 @@
-from palimpzest import Field
+from .elements import Field
```
mdr223, line 1: I've tried to avoid local imports, so I might just push a change to make this `from palimpzest.elements import Field`

src/palimpzest/generators/generators.py
```diff
@@ -139,18 +147,24 @@ def generate(self, context: str, question: str) -> GenerationOutput:
 
         # extract the log probabilities for the actual result(s) which are returned
         answer_log_probs = self._get_answer_log_probs(dspy_lm, pred.answer)
+        # TODO refactor this out in a method like in ImageTextGenerator?
```
mdr223, line 150: This is a good idea, I'll also apply this change and push it to your branch.

src/palimpzest/solver/solver.py
```diff
@@ -143,8 +149,72 @@ def _imageToEquation(candidate: DataRecord):
...
+
+                start_time = time.time()
+                xls = pd.ExcelFile(BytesIO(candidate.contents), engine='openpyxl')
+                api_stats = ApiStats(api_call_duration_secs=time.time() - start_time)
+
+                dr.number_sheets = len(xls.sheet_names)
+                dr.sheet_names = xls.sheet_names
+
+                if shouldProfile:
+                    candidate._stats[td.op_id] = InduceNonLLMStats()
```
mdr223, line 181: I think we should include the `api_stats` here, I'll make this change.

tests/biofabricDemo.py
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/env python3
+""" This scripts is a demo for the biofabric data integration.
+Make sure to run:
+python src/cli/cli_main.py reg --path testdata/biofabric-urls/ --name biofabric-urls
```
mdr223, line 4: Thank you for documenting this!

mdr223 (approved):
This LGTM; I'm about to push a small commit with the three minor changes I mentioned I would add in the comments, and then this is all set for you to merge. Great job!

PR #26, opened 2024-07-09
Top-Down Optimizer

mdr223:
UPDATE: see bottom of thread for newest comment(s)

WIP: Still need to add the plan pruning algorithm

Done:
1. Add tests + bugfixes for CostEstimator
2. Removed special functionality for creating sentinel plans
3. Add ability to enable/disable conventional queries
4. Re-organized model selection strategies into LLMFilter, LLMBondedConvert, LLMConventionalConvert
5. Split TokenReduced strategy into one for Bonded queries and one for Conventional queries
6. Enabled pass-through of `verbose` flag to convert and filter operators


src/palimpzest/dataclasses.py
```diff
@@ -48,16 +48,13 @@ def __add__(self, other: GenerationStats) -> GenerationStats:
         dct['model_name'] = self.model_name      
         return GenerationStats(**dct)
     
-    def __radd__(self, other: GenerationStats) -> GenerationStats:
-        return self
-
```
mdr223, line 53: (note: we have two copies of `__radd__`, I'm removing one of them here)

mdr223:
Will populate this thread with an up-to-date comment describing changes tomorrow AM.

mdr223:
Here is an overview of the major changes in this PR:

1. The creation of the Optimizer class, which replaces the LogicalPlanner and PhysicalPlanner
- The Optimizer is constructed with a Policy, a CostModel (previously CostEstimator), an OptimizationStrategy, and a set of hyperparameters (e.g. boolean flags that allow/disallow different optimizations)
- The Optimizer's optimize() method takes in a query plan (i.e. a user's PZ Dataset) and returns a list of physical plans, according to the given Policy we're optimizing for, the CostModel which produces estimates for each operator, and the OptimizationStrategy which dictates whether we only want the optimal plan, the set of possibly optimal plans (when considering error bounds), the set of sentinel plans, etc.
- The code for the Optimizer (and related classes, including Tasks, Rules, and optimization primitives such as Groups and Expressions) is in src/palimpzest/optimizer which effectively replaces src/palimpzest/planner
- This code -- specifically the Rules -- also implements the functionality of Strategies, which led to the removal of src/palimpzest/strategies; some code from src/palimpzest/strategies which specified convert operations was moved into new files in src/palimpzest/operators 
2. The CostModel (renamed from CostEstimator) now computes upper and lower bounds for the estimated statistics of interest, such as time_per_record, cost_per_record, selectivity (which is used to estimate cardinality), and quality
- The CostModel is initialized with sample execution data, which is used to compute per-operator estimates of the statistics of interest
- The CostModel's __call__ method takes in a physical operator and the OperatorCostEstimates from the physical operator's source operator and outputs a PlanCost object, which contains the total time, cost, and quality estimates (plus bounds) for this operator
    - NOTE: the OperatorCostEstimates and the new PlanCost dataclasses ended up being very similar -- the only difference is that OperatorCostEstimates is used to hold per-record estimates while the PlanCost is used to hold operator-level estimates. By changing the semantics of the dataclass(es) we may be able to combine these into a single dataclass.
- The code for this was moved out of src/palimpzest/execution into its own sub-directory src/palimpzest/cost_model
- Some future work will go into:
    - computing tighter lower and upper bounds
    - using something along the lines of matrix completion to better generalize our estimates from sample data to unseen physical operations
3. Replaced HardcodedConvert with NonLLMConvert; this enables programmers to write programs with lambda-style convert operations (very similar to how NonLLMFilter works for lambda-style filters)
- Code for NonLLMConvert is in src/palimpzest/operators/convert.py
- Changes were made to src/palimpzest/sets/sets.py to enable this syntax
- A Rule was implemented to perform pattern-matching and physical operator substitution for NonLLMConvert operations
4. Refactored src/palimpzest/execution/ into more modular classes
- ExecutionEngine is left with two unimplemented methods: execute() and execute_plan()
- PlanExecutors (which still inherit from the ExecutionEngine) implement the execute_plan() method; these handle details such as parallel vs. single-threaded execution of an individual plan
- The other ExecutionEngine sub-classes implement the execute() method; these handle details such as sentinel vs. non-sentinel execution of a (set of) plans
- Finally, sub-classes can inherit from one sub-class which implements execute_plan() and one sub-class which implements execute() to provide a fully-functional ExecutionEngine
5. Standardized the computation of execution ids, plan ids, logical operator ids (new), physical operator ids, expression ids (new), and group ids (new)
- The new implementations, along with implemented __hash__ and __eq__ methods allow for placing these classes directly into sets and dictionaries, which is useful at times
- The get_op_params methods for various physical operators were updated to ensure that identical physical operators' ids will match regardless of their input -- which is necessary for correctly applying CostModel estimates to an operator as the Optimizer moves the operator around the plan's logical tree
6. Simplified naiveCostEstimates() across all physical operators
- sub-classes may override the method to produce their own estimates
- some sub-classes (particularly for convert operations) now call super().naiveCostEstimates() to get the "vanilla" cost estimates and then apply small updates to the relevant estimates that they affect; this cuts down on a lot of code duplication and long functions
7. Updated Policy class to have three functions: constraint(), choose(), and (for now) get_primary_metric()
- constraint() returns a boolean indicating whether or not a given plan satisfies the policy's constraint; this can be used by the optimizer to more aggressively prune sub-plans (this type of pruning is still a TODO)
- choose() used to select the best plan from a set of physical plans, now it is used as a comparison function to determine whether the plan in the first argument is better than the plan in the second argument (according to the Policy)
- get_primary_metric() returns the string associated with the primary metric the policy is trying to optimize (e.g. "quality" if it is trying to maximize quality, "cost" / "time" if it is trying to minimize cost / runtime); this is used by the Optimizer to check the appropriate bounds when selecting the optimal (set of) plan(s)
    - In the future, we will need to rethink this if/when we support multi-objective optimization of the form e.g. minimize f(cost, quality) possibly subject to some constraint
8. Made appropriate changes to tests/pytest to get them passing
9. Removed structs from constants.py which became unnecessary after the refactor streamlined our hierarchy of operator classes; added struct for OptimizationStrategy and AggFunc
- ExecutionStrategy may be added back, but right now it is not needed as the Execute class takes an execution_engine as direct input
- QueryStrategy and CodingStrategy are now directly implied by the physical operator class
- PlanType was only used to set the num_samples argument for execute_plan(), which can easily be set by the ExecutionEngine sub-class now that we have separate classes for Sentinel vs. Non-Sentinel execution
- AggFunc struct removes the need for entire src/palimpzest/elements/aggregateFunction.py file
10. House-cleaning / reducing clutter:
- Added count() and average() as PZ syntax in src/palimpzest/sets/sets.py
- Removed LogicalPlan as the new Optimizer no longer needs it (it is now implicitly constructed with Groups and Expressions)
- Removed need for dataset_type in MarshalAndScanDataOp constructor
- Deleted Modal parallel sub-directory b/c I don’t think we’re using it anymore, and if someone really needs it they can pull it from the git history
- Deleted sampler sub-directory as it also is not being used
- Removed deserialize() function from classes because we are not using it anywhere
- Renamed uuid —> id (and parent_uuid —> parent_id) in DataRecord to be a bit more semantically correct
- In response to some feedback, removed class attributes from physical operators in favor of class parameterization
- Simplified physical operator constructors
- Various small bug-fixes

src/palimpzest/datasources/datasources.py
```diff
@@ -161,10 +161,9 @@ def __init__(self,
 
     def getItem(self, idx: int):
         filepath = self.filepaths[idx]
-        dr = DataRecord(self.schema, scan_idx=idx)
-        dr.filename = os.path.basename(filepath)
+        pdf_filename = os.path.basename(filepath)
         with open(filepath, "rb") as f:
-            dr.contents = f.read()
```
vitaglianog, line 167: If I recall correctly, we were hitting some errors in some demos since we expect the "contents" field to be set in any given object. We might also fix the demos that make that assumption though.

src/palimpzest/execution/execution_engine.py
```diff
@@ -0,0 +1,333 @@
...
+
+    def get_parallel_max_workers(self):
+        # for now, return the number of system CPUs;
+        # in the future, we may want to consider the models the user has access to
+        # and whether or not they will encounter rate-limits. If they will, we should
+        # set the max workers in a manner that is designed to avoid hitting them.
+        # Doing this "right" may require considering their logical, physical plan,
+        # and tier status with LLM providers. It may also be worth dynamically
+        # changing the max_workers in response to 429 errors.
+        return multiprocessing.cpu_count()
```
vitaglianog, line 112: Maybe just to be "nice" let's do 
``` return int( 0.8* multiprocessing.cpu_count() ) ```

File: src/palimpzest/execution/execution_engine.py [hunk 4]
```diff
@@ -0,0 +1,333 @@
...
+
+
+    def get_max_quality_plan_id(self, plans: List[PhysicalPlan]) -> str:
+        """
+        Return the plan_id for the plan with the highest quality in the list of plans.
+        """
+        max_quality_plan_id, max_quality = None, -1
+        for plan in plans:
```
vitaglianog, line 120: Here's an idea: can we store plans sorted by their quality? (This should also somewow implicitly sort by cost)
Therefore if we have a lot of plans we dont need a full for loop. But if we are not hitting significant bottlenecks maybe let's leave it as it is.

src/palimpzest/execution/execution_engine.py
```diff
@@ -0,0 +1,333 @@
...
+            # identify the plan with the highest quality in the set
+            max_quality_plan_id = self.get_max_quality_plan_id(plans)
+
+            # execute the set of plans for a fixed number of samples
+            new_execution_data, new_records, new_plan_stats = self.execute_plans(list(plans), max_quality_plan_id, self.num_samples)
+            records.extend(new_records)
+            plan_stats.extend(new_plan_stats)
+
+            # set still processing to False if there are no new records
+            if new_records == []:
```
vitaglianog, line 234: Is there a safeguard against having empty new_records e.g., due to not passing a filter vs. being the last batch of execution?

src/palimpzest/execution/execution_engine.py
```diff
@@ -0,0 +1,333 @@
...
+        plans = optimizer.optimize(dataset)
+        plan_set, prev_plan_set = set([plan.plan_id for plan in plans]), None
+        trials = 1
+        still_processing = True
+        while (
+            (trials < self.pareto_strategy_min_trials or plan_set != prev_plan_set)
+            and still_processing
+        ):
+            # identify the plan with the highest quality in the set
+            max_quality_plan_id = self.get_max_quality_plan_id(plans)
```
vitaglianog, line 281: Either this should be plan_set, or could be moved out of the while loop right?

vitaglianog: Or maybe I am a bit confused by the structure of the loop. We are always iterating on `plans` but keeping track of `plan_set` for the termination condition?
I think it would be good to add a short docstring of this function for future reference

src/palimpzest/execution/plan_executors/single_threaded_execution_engine.py
```diff
@@ -0,0 +1,299 @@
...
+    """
+
+    def execute_plan(self, plan: PhysicalPlan,
+                     num_samples: Union[int, float] = float("inf"),
+                     max_workers: Optional[int] = None):
+        """Initialize the stats and the execute the plan."""
+        if self.verbose:
+            print("----------------------")
+            print(f"PLAN[{plan.plan_id}] (n={num_samples}):")
+            plan.printPlan()
```
vitaglianog, line 30: Is there any difference between plan.printPlan() and print(plan) ? If so, maybe we should unify

src/palimpzest/datasources/datasources.py
mdr223: Demos seem to be working now, so I'll assume we are okay on this for now -- but happy to revisit if that turns out not to be the case

src/palimpzest/execution/execution_engine.py
mdr223: Good idea, will update

src/palimpzest/execution/execution_engine.py
mdr223: Good catch! Updating this logic to check the datasource length against the scan idx

src/palimpzest/execution/execution_engine.py
mdr223: Ended up scrapping this function entirely as this seems to be an optimization strategy which is not really useful

src/palimpzest/execution/execution_engine.py
mdr223: We can definitely do that -- right now the # of plans is very small so I will leave as-is, but definitely something we can do in a subsequent PR

src/palimpzest/execution/plan_executors/single_threaded_execution_engine.py
mdr223: `printPlan()` has slightly nicer syntax (the `inputSchema --> Operator --> outputSchema`) syntax; so I'll update `__str__` to be the body of `printPlan()`.

mikecafarella:
LGTM!!!!  (Kidding, we reviewed it at length.)

PR #30, opened 2024-10-20
Lint fix all fixable errors with ruff

ramipellumbi:
- Add ruff.toml to specify linting config. See: https://docs.astral.sh/ruff/editors/setup/ for setting Ruff up for your environment
- Alphabetize corelib fields
- Lint the following across `demos/`, `src/`, and `tests/`
  - Remove all unused imports
  - Sort all imports
  - Remove unused variable declarations
  - Fix `not x in y` to `x not in y`
  - Remove unnecessary usage of f-strings
  - Ruff: Do not call `setattr` with a constant attribute value. It is not any safer than normal property access.
- Modify .gitignore to ignore local [Zed](https://zed.dev/) configuration settings


mdr223 (commented):
These changes look good to me, although there are two minor details around:
1. `import context`, and
2. The choice of linter and inclusion of the ruff.toml file

Which I just want to make everyone has visibility on. Once we discuss those two minor issues then I'm happy to approve and merge.

ruff.toml
```diff
@@ -0,0 +1,9 @@
+# Config https://docs.astral.sh/ruff/configuration/
```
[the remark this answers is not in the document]
mdr223: Specifically, my questions in order are:
1. Do we have a preference on the choice of linter?
2. Does anyone have a concern with including the linter style/configuration file in the repo?

My personal answers are:
1. I have no preference on choice of linter, and am happy to try out Ruff
2. I am okay with including the linting style/config file as long as it is not something which other developers would be forced to use in their local development

My general belief is that it is good to have a consistent style / choice of linter -- however, I don't want to force it upon other developers if they'd prefer to develop in some other fashion. I know some folks who use their own linters for personal development, and then adhere to project-specific linting standards as a "final check" before pushing to a repo.

I believe that adding the `ruff.toml` file doesn't force a developer to use it -- it just serves as a declaration of the linting standards for the project. In this case, I'm okay with having it in the repo, although I might suggest moving it underneath a directory named `linter/` (but I will not die on that hill so feel free to ignore).

demos/bdf-suite.py
```diff
@@ -4,29 +4,24 @@
...
-import os
 
 pz.DataDirectory().clearCache(keep_registry=True)
 
 
+
+import matplotlib.pyplot as plt
 import networkx as nx
 import streamlit as st
-import context
```
vitaglianog, line 18: See above

demos/context.py
```diff
@@ -1,8 +1,7 @@
-import sys
 import os
+import sys
```
vitaglianog, line 2: I am genuinely curious if there is a rule/best practice for the order of these system imports?

ramipellumbi: In my experience, it is that they occur at the top of imports and are alphabetized within their import group

vitaglianog (approved):

demos/context.py
mdr223: I think the only one that most linters apply universally is alphabetical ordering. On my own I was sort of applying the following practice (which I picked up at my first job):
```
# project `from` imports first
from palimpzest.* import *

# project direct imports second
import palimpzest.* as *

# external from imports third
from abc import *

# external direct imports fourth
import pandas as pd 
```
But it looks like Ruff is just doing alphabetical, and `import` before `from`

demos/bdf-mimic.py
```diff
@@ -1,14 +1,13 @@
 # Note: include in the tests folder a .env file that contains the API keys for the services used in the tests
 import os
+
 if not os.environ.get('OPENAI_API_KEY'):
-    import context
+    pass
```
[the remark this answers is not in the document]
mdr223: I think a simple fix which will satisfy the linter (and my thoughts on moving this functionality out of an import) might be to just convert this into a utility function; @ramipellumbi would you like to take a stab at that? (If not I'm happy to do it)

ruff.toml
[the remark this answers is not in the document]
mdr223: @ramipellumbi are you opposed to moving `Ruff.toml` --> `linter/Ruff.toml`? If you'd prefer to leave it as is I won't object; I just try to leave the top-level as minimalist as possible in terms of non-directory files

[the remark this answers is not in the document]
ramipellumbi: Generally they should stay top level. 

Similar to [ESLint](https://eslint.org/docs/latest/use/configure/configuration-files#cascading-configuration-objects), Ruff supports hierarchical configuration, such that the "closest" config file in the directory hierarchy is used for every individual file, with all paths in the config file (e.g., exclude globs, src paths) being resolved relative to the directory containing that config file. If we place it in linter then nothing else will see it without manually passing in the config, which hides it from usage (which is not what you want?). If the plan is everyone bring their own linter (not recommended), then this is the approach for sure and I can do it. Either way, lmk

demos/context.py
ramipellumbi: Ruff is doing import groups and sorting within the groups. It is not just strictly alphabetical

demos/bdf-suite.py
ramipellumbi: Should this be reverted now or will the context handling happen in a follow up?

ramipellumbi: In addition to import fixes for demos, the demos rely on packages not present in the pip install (this is more obvious when using a venv for the installation rather than global python which users of might have installed). I think it is worth giving the demos an install and run script that create a .venv folder with the requirements for all demos installed into it and then run the relevant demo via something like `my_script --demo_name` (would make demos more easily reproducible).

That said, that should be a follow up if at all desired

ruff.toml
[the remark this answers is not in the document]
mdr223: This is reason enough to leave it as is, thanks for the explanation 👍

demos/bdf-suite.py
mdr223: I'm adding a function to do the environment loading, will push in a minute

mdr223:
Great job @ramipellumbi! Merging now.

PR #35, opened 2024-10-24
Make dynamic property assignment in `DataRecord` safer

ramipellumbi:
- Fix all lint errors in `elements/`
- Add `| None` typing to class values that initialize to `None` in `logical.py`
- Explicit pass of `None` in for `input_schema` on `CacheScan`

### DIFF
--- a/src/palimpzest/elements/filters.py
+++ b/src/palimpzest/elements/filters.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Any, Dict
+from typing import Any, Callable, Dict
 
 
 #############################
@@ -11,7 +11,7 @@ from typing import Any, Dict
 class Filter:
     """A filter that can be applied to a Set"""
 
-    def __init__(self, filterCondition: str | None = None, filterFn: callable | None = None) -> None:
+    def __init__(self, filterCondition: str | None = None, filterFn: Callable | None = None) -> None:
         self.filterCondition = filterCondition
         self.filterFn = filterFn
 
@@ -28,7 +28,7 @@ class Filter:
         # custom hash function
         return hash(self.filterCondition) if self.filterCondition is not None else hash(str(self.filterFn))
 
-    def __eq__(self, other: Filter) -> bool:
+    def __eq__(self, other) -> bool:
         # __eq__ should be defined for consistency with __hash__
         return (
             isinstance(other, Filter)
--- a/src/palimpzest/elements/groupbysig.py
+++ b/src/palimpzest/elements/groupbysig.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Any, Dict
+from typing import Any, Dict, Type
 
 from palimpzest.corelib import Field, OperatorDerivedSchema, Schema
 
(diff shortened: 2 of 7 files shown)


mdr223:
Pushed a small commit to:
- Apply a sanity check (to replace our previous setting of `kwargs['inputSchema'] = None`) to formally ensure that we never initialize `CacheScan` or `BaseScan` with a value other than `None`
- Give `inputSchema` a default setting of `None` in `LogicalOperator`. Without this, the call to the `BaseScan` constructor in the `Optimizer` would throw an exception. I considered just adding `inputSchema=None` to that constructor call, but I want to enforce correct behavior where possible.
- Heeded the advice of Ruff to remove `[]` as a default value in function signatures
- Made `schema` a property of the `Set` class; this has been a long-needed fix (kudos to @ramipellumbi for planting the `property` seed elsewhere) and is also needed for L142 in `optimizer.py`
- Fixed a fun bug 🐛 where the definition of `__setattr__` and `__getattr__` in `DataRecord` created an infinite recursion loop b/c when we initialized `self._data = {}` in the constructor, it would call `__setattr__(self, "_data", {})`. This would in turn call `__getattr__(self, "_data")` (from the `else` branch of `__setattr__`) which would then call itself (`__getattr__`) infinitely many times on the line `if name in self._data:`
- Removed `scan_idx` from list of class attributes in `__setattr__` b/c we really want it to be discarded after the constructor

ramipellumbi:
> Pushed a small commit to:
> 
>     * Apply a sanity check (to replace our previous setting of `kwargs['inputSchema'] = None`) to formally ensure that we never initialize `CacheScan` or `BaseScan` with a value other than `None`
> 
>     * Give `inputSchema` a default setting of `None` in `LogicalOperator`. Without this, the call to the `BaseScan` constructor in the `Optimizer` would throw an exception. I considered just adding `inputSchema=None` to that constructor call, but I want to enforce correct behavior where possible.
> 
>     * Heeded the advice of Ruff to remove `[]` as a default value in function signatures
> 
>     * Made `schema` a property of the `Set` class; this has been a long-needed fix (kudos to @ramipellumbi for planting the `property` seed elsewhere) and is also needed for L142 in `optimizer.py`
> 
>     * Fixed a fun bug 🐛 where the definition of `__setattr__` and `__getattr__` in `DataRecord` created an infinite recursion loop b/c when we initialized `self._data = {}` in the constructor, it would call `__setattr__(self, "_data", {})`. This would in turn call `__getattr__(self, "_data")` (from the `else` branch of `__setattr__`) which would then call itself (`__getattr__`) infinitely many times on the line `if name in self._data:`
> 
>     * Removed `scan_idx` from list of class attributes in `__setattr__` b/c we really want it to be discarded after the constructor

Dang, the recursion bug is a goofy mistake -- my bad

ramipellumbi:
> Pushed a small commit to:
> 
>     * Apply a sanity check (to replace our previous setting of `kwargs['inputSchema'] = None`) to formally ensure that we never initialize `CacheScan` or `BaseScan` with a value other than `None`
> 
>     * Give `inputSchema` a default setting of `None` in `LogicalOperator`. Without this, the call to the `BaseScan` constructor in the `Optimizer` would throw an exception. I considered just adding `inputSchema=None` to that constructor call, but I want to enforce correct behavior where possible.
> 
>     * Heeded the advice of Ruff to remove `[]` as a default value in function signatures
> 
>     * Made `schema` a property of the `Set` class; this has been a long-needed fix (kudos to @ramipellumbi for planting the `property` seed elsewhere) and is also needed for L142 in `optimizer.py`
> 
>     * Fixed a fun bug 🐛 where the definition of `__setattr__` and `__getattr__` in `DataRecord` created an infinite recursion loop b/c when we initialized `self._data = {}` in the constructor, it would call `__setattr__(self, "_data", {})`. This would in turn call `__getattr__(self, "_data")` (from the `else` branch of `__setattr__`) which would then call itself (`__getattr__`) infinitely many times on the line `if name in self._data:`
> 
>     * Removed `scan_idx` from list of class attributes in `__setattr__` b/c we really want it to be discarded after the constructor

The ruff lint point should be a public PSA. Python default arguments to functions are evaluated WHEN THE FUNCTION IS DEFINED NOT CALLED. So, if you have a mutable default argument like a list or set, that will be shared across all calls. Very bad. E.g.,

```
def append_to(element, to=[]):
    to.append(element)

my_list  = append_to(12) # [12]
my_other_list = append_to(42) # [12, 42]

```

mdr223 (approved):
Passing tests + demos

PR #36, opened 2024-10-24
Fix all linting issues in `dataclasses.py`

ramipellumbi:
Type Annotations:
* Updated `total_input_tokens` and `total_output_tokens` to `float` in `GenerationStats` and `RecordOpStats` classes. ([[1]](diffhunk://#diff-7e56f877be2f73c204dd6c6b2f36dece6b1ba5534ea1acf56b870d0b4ef2ac63L19-R24), [[2]](diffhunk://#diff-7e56f877be2f73c204dd6c6b2f36dece6b1ba5534ea1acf56b870d0b4ef2ac63L164-R171))
* Changed type annotations to use union types (e.g., `str | None`) for optional fields in multiple classes (`PlanStats`, `ExecutionStats`, `OperatorCostEstimates`, and `PlanCost`). ([[1]](diffhunk://#diff-7e56f877be2f73c204dd6c6b2f36dece6b1ba5534ea1acf56b870d0b4ef2ac63L263-R268), [[2]](diffhunk://#diff-7e56f877be2f73c204dd6c6b2f36dece6b1ba5534ea1acf56b870d0b4ef2ac63L318-R323), [[3]](diffhunk://#diff-7e56f877be2f73c204dd6c6b2f36dece6b1ba5534ea1acf56b870d0b4ef2ac63L361-R387), [[4]](diffhunk://#diff-7e56f877be2f73c204dd6c6b2f36dece6b1ba5534ea1acf56b870d0b4ef2ac63L418-R441))

Code Refactoring:
* Refactored type check for `record_op_stats_lst` to use `isinstance` for better readability (and lint fix) ([src/palimpzest/dataclasses.pyL225-R229](diffhunk://#diff-7e56f877be2f73c204dd6c6b2f36dece6b1ba5534ea1acf56b870d0b4ef2ac63L225-R229))
* Return `self` in `__iadd__` of `OperatorStats`. ([src/palimpzest/dataclasses.pyL240-R245](diffhunk://#diff-7e56f877be2f73c204dd6c6b2f36dece6b1ba5534ea1acf56b870d0b4ef2ac63L240-R245))

### DIFF
--- a/src/palimpzest/dataclasses.py
+++ b/src/palimpzest/dataclasses.py
@@ -16,10 +16,10 @@ class GenerationStats:
     # raw_answers: Optional[List[str]] = field(default_factory=list)
 
     # the total number of input tokens processed by this operator; None if this operation did not use an LLM
-    total_input_tokens: int = 0.0
+    total_input_tokens: float = 0.0
 
     # the total number of output tokens processed by this operator; None if this operation did not use an LLM
-    total_output_tokens: int = 0.0
+    total_output_tokens: float = 0.0
 
     # the total cost of processing the input tokens; None if this operation did not use an LLM
     total_input_cost: float = 0.0
@@ -161,10 +161,10 @@ class RecordOpStats:
     generated_fields: Optional[List[str]] = None
 
     # the total number of input tokens processed by this operator; None if this operation did not use an LLM
-    total_input_tokens: int = 0.0
+    total_input_tokens: float = 0.0
 
     # the total number of output tokens processed by this operator; None if this operation did not use an LLM
-    total_output_tokens: int = 0.0
+    total_output_tokens: float = 0.0
 
     # the total cost of processing the input tokens; None if this operation did not use an LLM
     total_input_cost: float = 0.0
@@ -222,7 +222,7 @@ class OperatorStats:
         plan_id: str,
     ):
         # convert individual record into list
-        if type(record_op_stats_lst) != type([]):
+        if not isinstance(record_op_stats_lst, list):
             record_op_stats_lst = [record_op_stats_lst]
 
         # update op stats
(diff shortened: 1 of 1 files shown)


src/palimpzest/dataclasses.py
```diff
@@ -16,10 +16,10 @@ class GenerationStats:
     # raw_answers: Optional[List[str]] = field(default_factory=list)
 
     # the total number of input tokens processed by this operator; None if this operation did not use an LLM
-    total_input_tokens: int = 0.0
+    total_input_tokens: int = 0
```
mdr223, line 19: One subtle note on this, I think what we want to do is to change the type here (and in `RecordOpStats`) to `float` and add a second comment between L18-L19 to the effect of: `# total_input_tokens is typed as a float because GenerationStats may be amortized (i.e. divided) across a number of output records` (same goes for `total_output_tokens`)

src/palimpzest/dataclasses.py
```diff
@@ -233,11 +233,17 @@ def add_record_op_stats(
             self.total_op_time += record_op_stats.time_per_record
             self.total_op_cost += record_op_stats.cost_per_record
 
-    def __iadd__(self, op_stats: OperatorStats):
+    def __add__(self, op_stats: OperatorStats):
```
mdr223, line 236: I could be mistaken, but I believe `__iadd__` provides support for the `+=` operator, whereas `__add__` provides support for the `+` operator. Thus, if we want to make this change, we would also need to update every instance of:
```
record_op_stats += other_record_op_stats
```
To be:
```
record_op_stats = record_op_stats + other_record_op_stats
```
Alternatively, we could ignore this linting error in the `ruff.toml` (or perhaps there's a third option which I'm not thinking of?). Overall, I don't have a firm stance on this, other than my belief that support for the `+=` syntax is nice b/c it makes code a tiny bit more readable elsewhere, but I'm happy to switch every `+=` to `+` if you'd prefer. LMK.

mdr223: (I see that you made an update below to the effect of `op_stats = op_stats + other_op_stats`, I think there might be a few places in `operators/` where this change also needs to take effect -- but I'll confirm if we decide to deprecate `+=`.

mdr223 (commented):
Left a comment re: the `__iadd__` operation. Let me know what you think and in the meantime I'll see if I can locate other instances of using this operation.

src/palimpzest/dataclasses.py
ramipellumbi: Reverted to `__iadd__` and fixed the issue (was not returning `self`)

src/palimpzest/dataclasses.py
ramipellumbi: Resolved

mdr223 (approved):

PR #39, opened 2024-10-24
Fix default mutable args

### LINKED ISSUE #38: Default argument cleanup for mutable default arguments
The below excerpt is a modified version of a write up done by [Blue9](https://github.com/blue9/).

This is non-obvious behavior: In Python default arguments to functions are evaluated once when the function is defined (as a global), not called. So if you have a mutable default argument like a list, dictionary, or set, that will be shared across all calls. Example (from [here](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments)):

```
def append_to(element, to=[]):
    to.append(element)
    return to

my_list = append_to(12)
print(my_list)  # prints [12]

my_other_list = append_to(42)
print(my_other_list)  # prints [12, 42]
```

It is better to avoid mutable default arguments if possible (even if you're not mutating anything). Better to default to `None` and set the default within the class or function in the event it is `None`.

Small Note: Pydantic handles this differently ([docs](https://docs.pydantic.dev/latest/concepts/models/#fields-with-non-hashable-default-values)) and does allow you to specify mutable default arguments as it will deep-copy them every time the model is created.

The newly added ruff linter will call this out as an issue, too. It exists in more places but `optimizer/tasks.py`  is one of the more pertinent violations.

ramipellumbi:
- This PR will close #38 and all other mutable default argument sets
- Removes unnecessary enumerate calls
- Remove unused vars in `generators.py`

### DIFF
--- a/src/palimpzest/corelib/schemas.py
+++ b/src/palimpzest/corelib/schemas.py
@@ -217,7 +217,7 @@ class Table(Schema):
         # Take the rows in the record_dict and turn them into comma separated strings
         rows = []
         # only sample the first MAX_ROWS
-        for i, row in enumerate(record_dict["rows"][:MAX_ROWS]):
+        for row in record_dict["rows"][:MAX_ROWS]:
             rows += [",".join(map(str, row)) + "\n"]
         record_dict["rows"] = rows
         header = ",".join(record_dict["header"])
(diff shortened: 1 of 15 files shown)


mdr223 (approved):
Nice job! All changes LGTM.

PR #42, opened 2024-10-24
Assert DataDirectory.current_config is defined on access. udfs cleanup.

ramipellumbi:
- Remove root import in udfs.py


mdr223 (approved):
All good changes.

PR #43, opened 2024-10-26
Remove import palimpzest intra package

ramipellumbi:
In local project files, the desired imports should be direct imports. Long term, the package root `__init__` should export only the desired public exports for users of the library rather than absolutely everything. This PR removes all instances of importing from the root in `src/`


mdr223 (approved):
Thanks for cleaning this up!

PR #49, opened 2024-12-12
Dynamic schema

vitaglianog:
This branch introduces a new module, called `SchemaBuilder`. The purpose of this module is to allow users to create palimpzest schemas dynamically from existing files.
The base syntax is the following 
```
schema = pz.SchemaBuilder.from_file(schema_path)
```

The builder can take as input a custom name, set of attributes to include, description and a type. 
It currently supports four file types: JSON, YAML, JSON-LD, and CSV.
- The most powerful readers are JSON and YAML, where users can fully specify information such as the parent class of the schema and the type of the inputs.
- The CSV reader is a convenience function that takes an existing csv and generates a schema that matches attribute names, and heuristically based on pandas detection of column types it casts the pz fields to different native field types.
- The JSON-LD reader is used to read schema files specified through json-ld formats (such as the ones distributed within synapse).

The module uses a fairly naive logic (the extension of the file) to check which function to call.


pyproject.toml
```diff
@@ -78,6 +79,7 @@ namespaces = false
 
 
 [tool.pytest.ini_options]
+testpaths = ["tests/pytest"]
```
vitaglianog, line 82: This is used to call pytest from root dir and avoid the warning about `conftest.py`

testdata/images-tiny.zip
mdr223, line 1: Is this meant to be included in the commit?

mdr223 (approved):
Left two minor comments; pre-approving as I think you can handle those and merge when ready.

testdata/images-tiny.zip
vitaglianog: Oops nice catch!

PR #54, opened 2025-01-06
New package branch

mikecafarella:
I moved around packages a bit.  It's not the absolute most beautiful layout yet, but it has a couple of advantages:
1) It puts most of the basic things into a new top-level directory called core
2) It puts most of the query-processing code into a new top-level directory called query
3) It changes what gets loaded automatically when doing import palimpzest

As a result, it is possible to load a lot less when doing simple operations. In particular, doing ls-data was infamously slow because any "import palimpzest" took almost 3 seconds.

Now it's a lot better.  You can do ls-data and "import palimpzest" in under a second.  There's still some code in DataDirectory that's too slow, but this is a step in the right direction.


src/palimpzest/policy.py
mdr223, line 1: I would move this file into `core` as well if we're moving `dataclasses.py` (and others) out of `src/palimpzest/`

src/palimpzest/query/__init__.py
```diff
@@ -0,0 +1,31 @@
+from palimpzest.query.execution.execute import Execute
```
mdr223, line 1: Same note from before re: being careful about circular import hell with these intermediate `__init__.py` files.

src/palimpzest/query/optimizer/optimizer.py
```diff
@@ -16,14 +16,14 @@
     LogicalOperator,
     RetrieveScan,
 )
-from palimpzest.optimizer import (
+from ..optimizer import (
```
mdr223, line 19: I would like to eliminate any and all relative imports in favor of absolute imports if possible (these can mask circular import issues).

I.e. I would instead use:
```
from palimpzest.query.optimizer import (
    IMPLEMENTATION_RULES,
    TRANSFORMATION_RULES,
)
```

tests/pytest/test_dynamicschema.py
```diff
@@ -1,21 +1,22 @@
...
 import palimpzest as pz
+import palimpzest.schema_builder as sb
 from palimpzest.constants import Model
 
 data_path = "tests/pytest/data/"
 
 
 def test_dynamicschema_jsonld():
-    clinical_schema = pz.SchemaBuilder.from_file(data_path + "synapse_schema.jsonld", schema_type=pz.TextFile)
+    clinical_schema = pz.sb.SchemaBuilder.from_file(data_path + "synapse_schema.jsonld", schema_type=pz.TextFile)
```
mdr223, line 11: With the new import on line 4, would this now just become `sb.SchemaBuilder...`?

mdr223 (commented):
Left a few comments, mostly small ones but some meta / high-level ones as well. Will approve once I get a chance to run unit tests on this branch.

demos/simpleDemo.py
```diff
@@ -18,22 +18,22 @@
...
 import palimpzest as pz
 from palimpzest.constants import Cardinality
-from palimpzest.elements.groupbysig import GroupBySig
-from palimpzest.elements.records import DataRecord
+from palimpzest.core.elements.groupbysig import GroupBySig
+from palimpzest.core.elements.records import DataRecord
 
 
-class ScientificPaper(pz.PDFFile):
+class ScientificPaper(pz.core.PDFFile):
```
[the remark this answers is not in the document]
mikecafarella: This is a good comment, but let's separate style and nice package design commits from the near-term goal of a crude reorg to help with latency.

demos/simpleDemo.py
```diff
@@ -367,12 +375,13 @@ def print_table(records, cols=None, gradio=False, plan_str=None):
 
     execution_engine = None
     executor = args.executor
+    import palimpzest.query
```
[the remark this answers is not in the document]
mikecafarella: I've made changes that make this line obsolete

demos/simpleDemo.py
```diff
@@ -367,12 +375,13 @@ def print_table(records, cols=None, gradio=False, plan_str=None):
...
     if executor == "sequential":
-        execution_engine = pz.NoSentinelSequentialSingleThreadExecution
+        execution_engine = pz.query.NoSentinelSequentialSingleThreadExecution
```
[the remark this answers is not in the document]
mikecafarella: Good point, but again let's separate style mods from near term goal.

PR #57, opened 2025-01-09
New features

chjuncn:
1. fix imports for pytest
2. centralize hash function for ids
3. add to_df, from_df for DataRecord, so that it'll be easy for users to plug in or pull out data to or from PZ pipelines.

Next step, I'll think about how to make df more natural through the pipelines.


mdr223 (approved):
Changes look good to me! I especially like the ability to create DataRecords directly from a DataFrame, and thank you for making a util for the hash function!

chjuncn:
Thanks Matt for your review!

I uninstalled backports.tarfile, and the program works well today... I
don't know what happened yesterday, so never mind :)

PR #63, opened 2025-01-17
Quickstart demo

vitaglianog:
This PR contains code to run a quickstart demo. This is provided as a jupyter notebook within the project root. A few notable changes to the repo:
 - The `Execute()` function now takes a default of `nocache=True` and raises an Exception if it is set to `False`.
 - The `enron-tiny` folder has been added to testdata to make it easier to run the quickstart notebook. I am happy if we find a different solution but this was the least effort one for people to run the notebook.
 - The Datarecord `repr` is now the same as the `str` method


mikecafarella (commented):
The narrative is clear and I tried it with success.  I think this is as about as succinct as one can get PZ right now, without our planned changes to syntax. I wish it were smaller, but that's for a different commit.

PR #64, opened 2025-01-18
Fix up the pyproject.toml to avoid unnecessary builds from source

mikecafarella:
A few libraries, notably pyarrow, take a long time to build. We added version rules to encourage pip to install these from prebuilt wheels, not compile from source


mdr223 (approved):
LGTM

PR #68, opened 2025-01-22
Support list, df when init Dataset()

chjuncn:
Support list, df when init Dataset()

1. Auto generate dataset_id for memory source
2. create _tempRegistry for  memory data, it's temporary and won't be saved to disk. This is for providing unified interface for Datasource, we will need to revisit how to provide the best cache story in the system.
3. Mike mentioned Dataset init() can only take str/DataSource and move the logic to Datasource, I moved some logic to datamanager, and I still make init() takes list/str/df/datasource, since users need to create DataSource() when creating Dataset() if I don't do this, I'm in less favor of that. But if you still think that's better, I can update it. Or after I submit this change, Mike can update the code based on the current base.


mdr223 (approved):
LGTM

PR #78, opened 2025-01-27
1. support add_columns in Dataset; 2. support run().to_df(); 3. add demo in df-newinterface.py

chjuncn:
Support workflow:

df = pd.read_csv("testdata/enron-tiny.csv")
qr2 = pz.Dataset(df)
qr2 = qr2.add_columns({"sender": ("The email address of the sender", "string"), 
                        "subject": ("The subject of the email", "string"),#
                        "date": ("The date the email was sent", "string")})

qr3 = qr2.filter("It is an email").filter("It has Vacation in the subject")

output_df = qr3.run().to_df()


src/palimpzest/core/lib/schemas.py
```diff
@@ -274,6 +274,48 @@ def from_df(df: pd.DataFrame) -> Schema:
...
+        new_field_names = list(cls.field_names())
+        new_field_types = list(cls.field_map().values())
+        new_field_descs = [field._desc for field in new_field_types]
+
+        # Add new fields, skip existing fields
+        for field_name, field_desc in fields.items():
+            if field_name in new_field_names:
+                continue
+            new_field_names.append(field_name)
+            new_field_types.append(StringField(desc=field_desc))  # Assuming StringField for new fields
```
mdr223, line 305: For now let's leave this as-is, but can you add a #TODO here for us to revisit this in #84? In the PR 
 for that issue, we will have users also provide the field type which we should apply here.

src/palimpzest/sets.py
```diff
@@ -212,6 +212,11 @@ def convert(
             desc=desc,
             nocache=self._nocache,
         )
+    
+    # This is a convenience for users who like DataFrames-like syntax.   
+    def add_columns(self, columns:dict[str, str]) -> Dataset:
+        output_schema = self.schema.add_fields(columns)
+        return self.convert(output_schema, udf=None, cardinality=Cardinality.ONE_TO_MANY, depends_on=None, desc="Add columns " + str(columns))
```
mdr223, line 219: I think `cardinality` needs to be an argument to this function the same way it is for `convert()` and then that should be passed into `self.convert()` here.

`Cardinality.ONE_TO_MANY` does not refer to the number of columns generated by the convert, but the number of rows (i.e. records) generated for a given input record.

testdata/convert_enron_to_csv.py
mdr223, line 1: While I recognize the utility of having this code around, I would not commit this file to this repo.

I'm eventually supposed to move our demos and demo-related code to a new repository: https://github.com/mitdbg/pz-cookbook (if you cannot access this, let me know and I'll add your Github ID).

I think this file would belong there, in a `utils/` folder of sorts. Alternatively, see my comment on the next file for a suggestion of how you could bundle this file into the repo without committing it.

testdata/enron-tiny.csv
mdr223, line 1: Rather than committing our data files to Github we've decided to have the files automatically downloaded by the `testdata/download-testdata.sh` script.

If you can send this file to @vitaglianog via Slack, he can upload it to his CSAIL directory on your behalf.

demos/bdf-suite.py
```diff
@@ -220,24 +220,27 @@ def extract_references(processing_strategy, execution_strategy, optimizer_strate
...
     )
-    iterable = output.run(config)
-
+    data_record_collection = output.run(config)
+    
     references = []
     statistics = []
 
-    for idx, (reference, plan, stats) in enumerate(iterable):
+    for idx, record_collection in enumerate(data_record_collection):
```
mdr223, line 228: Is this correct? I.e. will `output.run(config)` output a list of `data_record_collection` objects? or just a single one?

demos/bdf-usecase3.py
```diff
@@ -75,11 +76,15 @@ def run_workload():
 
     tables = []
     statistics = []
-    for table, plan, stats in iterable:  # noqa: B007
+    for data_record_collection in iterable:  # noqa: B007
```
mdr223, line 79: Similar question here re: whether `iterable` will actually  be a list of `DataRecordCollection` objects?

demos/fever-demo.py
```diff
@@ -306,7 +306,7 @@ def get_item(self, idx: int, val: bool=False, include_label: bool=False):
 )
 
 record_jsons = []
-for record in records:
+for record in data_record_collection.data_records:
```
mdr223, line 309: I'm leaving a comment on this here, but I think this pattern exists in a few places.

If you think it's clean syntax, I might suggest that we add an `__iter__` method to `DataRecordCollection` which is implemented as:
```
def __iter__(self):
    yield from self.data_records
```
This way we can simply write:
```
for record in data_record_collection:
    ...
```
Just an idea, feel free to ignore it though if you're not a fan!

mdr223 (approved):
I've left a few comments:
- A few need to be addressed before merging (e.g. the comment re: `ONE_TO_MANY` cardinality in `add_columns()`)
- Some others are more style suggestions (or questions) which you can ignore if you prefer

I'm pre-emptively approving this PR so that once you make the changes in your timezone you can merge the PR without needing to wait for me to wake up :)

src/palimpzest/core/lib/schemas.py
chjuncn: added todo

src/palimpzest/sets.py
chjuncn: Make sense.
I'll make ONE_TO_MANY as default for add_columns

testdata/convert_enron_to_csv.py
chjuncn: SG, I'll remove this file and submit to the another repo.

demos/fever-demo.py
chjuncn: I like this idea :)

demos/bdf-usecase3.py
chjuncn: discussed offline. I'll add __iter__ in DataRecordCollection in the current PR, and submitted https://github.com/mitdbg/palimpzest/issues/87 for later.

demos/bdf-suite.py
chjuncn: commented in [demos/bdf-usecase3.py]

PR #92, opened 2025-01-29
Multi-LLM Refinement Pipeline for Query Output Validation 

yash94404:
- Implemented multi-LLM refinement pipeline for query output validation
- Added critic LLM evaluation step to assess Palimpzest-generated query output
- Introduced a second LLM stage that refines output based on critic feedback
- Added logging and debugging utilities for tracking LLM decision paths


mdr223:
Hi @yash94404, thanks for contributing this PR -- I look forward to shepherding it into `main` with you!

One quick note is that I recently added an initial set of docs which is deployed from the `gh-pages` branch and which can be found at [palimpzest.org](palimpzest.org).

I see that you've added a CI pipeline to deploy an `MkDocs` setup, but I just want to make sure that we don't overwrite the existing set of documentation until we get a chance to review your code together.

Overall, I'm excited to see this work, and will aim to give it a review this week!

mdr223:
Hi Yash, to summarize from our conversation, I think you'll want to make the following changes:

0. Move `docs/` changes into another branch and target this PR to `dev`
1. Create a new file in `src/palimpzest/query/operators/` which will hold the implementation of your physical operator 
    a. (e.g. `src/palimpzest/query/operators/critique_and_refine_convert.py` or something like that)
    b. I would suggest copying `mixture_of_agents_convert.py` to give you a starting point since it is somewhat similar
2. Implement your class in the new file and update the boilerplate methods as needed.
    a. Don't worry too much about updating `naive_cost_estimates()` as this function is a bit wonky.
3. Implement the `convert()` method for your class;
    a. Like the convert method in `MixtureOfAgentsConvert`, you will want to invoke a `generator` object multiple times: once to get an initial response, a second time to generate a critique, and a third time to get a refined output
    b. In order to enable the `generator` to use the correct prompt(s) you added in `prompts.py`, you will need to add two new entries in the `PromptStrategy` enum in `src/palimpzest/constants.py`
        i. One for the critique (e.g. `PromptStrategy.CRITIQUE`)
        II. And one for the refinement (e.g. `PromptStrategy.REFINE`)
4. Update `BaseGenerator._generate_developer_prompt()` and `BaseGenerator._generate_user_prompt()` to use the appropriate prompt depending on `self.prompt_strategy`
5. Finally, add a new rule in `src/palimpzest/query/optimizer/rules.py` which will enable the `Optimizer` to use your convert operator
    a. Once again, it might be a good idea to copy the `MixtureOfAgentsConvertRule` as a good place to get started

Please feel free to ping me on Slack with any questions or thoughts! I'm more than happy to help marshal this all the way into `main`!

Issue #84, opened 2025-01-27
Update Syntax to Reflect New Design Goals


mdr223:
Following today's design discussion, we would like to make the following changes to the PZ syntax:

1. split `.convert()` into --> `.sem_add_columns()` and `.add_columns()` (where the first one accepts a string description and the second accepts a udf)
2. split `.filter()` into --> `.sem_filter()` and `.filter()` (where the first one accepts a string filter, and the second accepts a udf)
3. change `.sem_add_columns()` to support the following syntax:
```
.sem_add_columns(
  cols=[
    {"name": "sender", "type": "string", "description": "The email address of the sender"},
    {"name": "subject", "type": "string", "description": "The subject of the email"},
    {"name": "date", "type": "string", "description": "The date the email was sent"},
  ]
)
```
We will also need to support UDF execution for `.add_columns()`:
```
.add_columns(
  cols=[
    {"name": "sender", "type": "string", "udf": compute_sender},
    ...
  ]
)
```
We should also support users who have an expensive udf that they only want to execute once and which computes all fields, e.g.:
```
.add_columns(
  udf=my_expensive_udf, types={"subject": "string", "sender": "string", "date": "string"}
)
```
In this^ example, `types` must map the field names to their types.

The implementations of `.sem_filter()` and `.filter()` should be simpler (the first accepts a `str` filter and the second accepts a `callable` filter)

PR #101, opened 2025-02-01
Code update for https://github.com/mitdbg/palimpzest/issues/84

chjuncn:
This implementation basically resolves https://github.com/mitdbg/palimpzest/issues/84.

One implementation is different from the #84:
.add_columns(
  cols=[
    {"name": "sender", "type": "string", "udf": compute_sender},
    ...
  ]
)

If add_columns() uses cols, udf, types as params, and cols supports "udf" and possibly "desc", it will make this function confusing again, as it has totally different ways to use this function. 

Instead, we just support udf and types for add_columns. If users need to specify different udfs for different columns, they should just call add_columns() multiple times for different columns.


mdr223 (approved):
Made some minor changes, mostly just after I realized we should probably re-use Python's built-in types (rather than mapping strings as I had suggested). Otherwise, this LGTM!

PR #103, opened 2025-02-03
Simplify Datasource

mdr223:
This PR covers changes which simplify the use of the `DataSource` class for:
1. Users writing programs (especially with validation data), and
2. The internal management of `DataSource`s in PZ

The primary changes I made were to:
1. Reduce the three data source classes (`DataSource`, `AbstractDataSource`, and `ValidationDataSource`) down to a single `DataSource` class
2. Limit the number of methods the user needs to implement to just `__len__()` and `get_item()`

This in turn led to a number of other changes / simplifications in the code including:
1. Using `datasource` directly in the scan operators (both logical and physical) instead of `dataset_id`
2. Changing `source_id` --> `source_idx` everywhere (this eliminated some weird renaming logic)
3. And various other small fixes which resulted from these changes

Finally, I added a large set of documentation for the `DataSource` class(es)


mikecafarella (commented):
I think the change is extensive but conceptually not that difficult and I generally like it. But I think the validation datasource example is too complicated for non-internals programmers to handle. We should probably have a new DataSource subclass called ValidationSource that helps them construct it correctly.

src/palimpzest/core/data/datasources.py
```diff
@@ -29,69 +27,164 @@
...
+            dict: A dictionary representing the item at the given index. The dictionary
+            keys (i.e. fields) should match the fields specified in the schema of the
+            data source, and the values should be the values associated with those fields.
+
+            .. code-block:: python
+
+                # Example return value
+                {"field1": value1, "field2": value2, ...}
+
+            If the data source is a validation data source, then the dictionary should
```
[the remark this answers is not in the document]
mdr223: To summarize my follow-up in Slack:

This approach is modeled after PyTorch, where dictionaries are the way users provide inputs and labels to the system. This tight-coupling makes it a bit easier for the programmer to reason about the validation, because the input and its label are provided within the same return object (i.e. dictionary).

Previously, we had a different approach where the user provided inputs in one datasource and labels in another. But this was painful, b/c it required the user to coordinate `source_id`s correctly across the two datasources. It also made custom scoring functions a bit wacky, as users needed to provide them in the `ValidationDataSource` while understanding that the computed value(s) would come from processing data in the non-validation `DataSource`.

Finally, one reason I am opposed to having a separate class for `ValidationDataSource` is that users will need to swap out their `DataSource` for a `ValidationDataSource` if they decide to add validation labels.

For example, imagine the human-in-the-loop setting where you code up a PZ pipeline, process the first 5 records, and inspect the outputs. A natural desire might be to examine the outputs, figure out what the labels should be, and then add those back into your datasource. Right now, this would just involve adding a `"labels"` key, and optionally a `"score_fn"` key (if you needed to score something other than exact matching). In the world where we have two types of datasources, you would need to change your `DataSource` to a `ValidationDataSource` and make any other changes to support this transition.

One thing I can do is add assertions in the sentinel query processors which perform a check to make sure that each output from the `val_datasource` has a `"labels"` key present. This can then throw an exception if no key is present, with an error message pointing the user to our validation documentation?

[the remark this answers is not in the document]
mdr223: Of course there can be many other (better) solutions than what I've outlined above.

This^ solution also does not address the case where validation labels need to be computed at the dataset level (e.g. for a `join`) rather than on a per-input basis.

But my overarching goal is to:
1) make it as simple as possible for users to provide validation labels
2) make it as clear as possible what the validation label(s) are for a given input (or dataset)

My primary concern with having a separate `ValidationDataSource` is that it forces users to reason about a potentially different class, when they probably just want to add labels to their existing `DataSource`.

sivaprasadsudhir:
I am okay with having a single DataSource class. But the current version of get_item() is too complicated to implement.

A few thoughts:
1. score_fn is a property of a field in the schema. It need not be set on a per-input basis. It should not be in the get_item() function. It should be set in a separate function.
2. We can support a per-input score_fn that compares ouput_record and target_record. But we would also need a score_fn that scores based on the output of the whole validation data.
3. It's unclear which indexes for get_item() have labeled examples that can be used for validation. 
4. How about the users implementing a separate function get_validation_items() that returns, all input records, corresponding output labels, and optional scoring functions?

mdr223:
> I am okay with having a single DataSource class. But the current version of get_item() is too complicated to implement.
> 
> A few thoughts:
> 
> 1. score_fn is a property of a field in the schema. It need not be set on a per-input basis. It should not be in the get_item() function. It should be set in a separate function.
> 2. We can support a per-input score_fn that compares ouput_record and target_record. But we would also need a score_fn that scores based on the output of the whole validation data.
I agree with both of these^ points, and can make changes to that effect

> 3. It's unclear to me which indexes for get_item() have labeled examples that can be used for validation.
In this paradigm, the user provides two datasources. Consider the following code, which I've modified slightly for the purpose of illustration:
```
# create and register data source
datasource = BiodexSource(dataset_id=dataset_id, split="test", ...)

# create and register validation data source
val_datasource = BiodexSource(dataset_id=val_dataset_id, split="val")

# load index
index_path = ".ragatouille/colbert/indexes/reaction-terms"
index = RAGPretrainedModel.from_index(index_path)

# construct plan
plan = Dataset(datasetid)
plan = plan.sem_add_columns(biodex_reactions_cols, depends_on=["title", "abstract", "fulltext"])
...

config = QueryProcessorConfig(
    policy=policy,
    val_datasource=val_datasource,
    execution_strategy="sentinel",
)
output = plan.run(config)
```
The datasource which is used to optimize the `plan` is the `val_datasource`, and in this datasource **every output must have a label**. Once the optimization is finished, the plan processes the `datasource`, which has no labels.

This is very similar to training vs. deploying NN models in PyTorch. Every element of your training/validation datasets must have a label, but when you deploy the model for inference you are only given data inputs.

> 4. How about the users implementing a separate function get_validation_items() that returns, all input records, corresponding output labels, and optional scoring functions?
Given my answer to (3.), I think you will see that this^ is effectively what `get_item()` does for a `DataSource` which is used for validation :)

mdr223:
To summarize ideas from my discussion w/Siva:
1. add a `get_label(idx)` function to `DataSource` which the user can implement for validation datasources; remove labels and scoring functions from `get_item(idx)`
2. add a `score_output(output, expected_output)` to `DataSource`, which the user can implement to provide feedback
3. construct a `ValidationExamples` class, which a user constructs with a list of dictionaries representing the expected output(s) for the execution of the plan on the (validation) datasource
4. **divorce the datasource from the plan** (renaming `Dataset` --> `Plan`)
5. add `datasource` as an input to `plan.run()`

An example of PZ code with human-in-the-loop validation would now look like the following:
```python
from palimpzest import DataSource, Plan, QueryProcessorConfig
import os

class EnronDataSource(DataSource):
    def __init__(self, email_cols, email_dir, num_examples = None):
        super().__init__(email_cols)
        self.email_fps = [os.path.join(email_dir, email) for email in os.listdir(email_dir)]
        if num_examples:
            self.email_fps = self.email_fps[:num_examples]

    def __len__(self):
        return len(self.emails)

   def score_output(output, expected_output):
       num_correct = sum([
           value == expected_output[key]
           for key, value in output.items()
            if key in expected_output
       ])
       return num_correct / len(expected_output)

   def get_label(self, idx):
        email = self.get_item(idx)
        sender = re.match("Sender: (.*?)\n", email)
        subject = re.match("Subject: (.*?)\n", email)

        return {"sender": sender, "subject": subject}

    def get_item(self, idx):
        email_fp = self.email_fps[idx]
        email = {
            "filename": email_fp,
            "contents": open(email_fp, 'r').read()
        }

        return email


# construct datasources
email_cols = [
    {"name": "sender", "type": str, "desc": "the sender of the email"},
    {"name": "subject", "type": str, "desc": "the subject of the email"},
]
val_datasource = DataSource(email_cols, "enron-emails/", num_examples=5)
full_datasource = DataSource(email_cols, "enron-emails/")

# construct plan
plan = Plan()
plan = plan.sem_add_columns(email_cols)
plan = plan.sem_filter("is about fraud")
plan = plan.sem_filter("is not a news article")

# run plan on validation data
val_output = plan.run(val_datasource, config=QueryProcessorConfig())
print(val_output.to_df())
```
Now the human gets to see the `val_output` and can use it as a starting point for their labelling:
```
# construct labels
val_records = [
    {"source_idx": 0, "sender": "henry@enron.edu", "subject": "blah"},
    {...},
]
val_examples = ValidationExamples(val_records)

# re-run plan with validation examples
new_output = plan.run(val_datasource, config=QueryProcessorConfig(val_examples))
```
Now the user is free to inspect their `new_output`. If they are satisfied with it, then can then execute:
```
final_output = plan.run(full_datasource, config=QueryProcessorConfig(policy="maxquality"))
```
The one piece which is missing from the example above is how we use the execution stats to make sure that `plan` is updated every time we call `plan.run()`. I'm open to suggestions on that.

mdr223:
More ideas:
1. `plan.optimize()` and `plan.run()`
2. store physical plan, optimizer stats, and validation examples in `plan`
3. rename `plan` to something else :D

mdr223:
## Summary of PR changes

**Note**: I did not change anything related to `val_datasource` (including tangential functions like `Dataset._set_data_source()`) as that will all be modified in a subsequent PR to reflect our discussion re: validation data.

**Second Note:** I have completely commented out `datamanager.py` and `config.py`; for now I am willing to leave the code around in case we desperately need it for PalimpChat. However, my hope is that PalimpChat can be tweaked to work without the data manager and those files can be deleted before merging `dev` into `main`

- Collapsed all four `DataSource` classes down to a single `DataReader` class
- Provided `DataReader` directly to scan operators (also renamed `DataSourcePhysicalOp` --> `ScanPhysicalOp`
- Switched from using `get_item()` --> `__getitem__()` in `DataReader`
- Removed `DataDirectory()` from `src/` entirely; this included commenting out things which made use of the cache (e.g. caching computed `DataRecords` and codegen examples)
- Got rid of `dataset_id` everywhere (which tracks with the previous bullet)
- Removed the `Config` class which was a relic of a bygone era (and also intertwined with the `DataDirectory()`)
- Updated all demos to use `import palimpzest as pz` to make the import statement(s) more welcoming
- Fixed one bug resulting from converts now producing union schemas. Instead of including the `output_schema` in an operators' `get_id_params()` we simply report the `generated_fields`.

src/palimpzest/sets.py
```diff
@@ -129,52 +108,33 @@ def json_schema(self):
...
-            else:
-                schema = Schema.from_df(source) if isinstance(source, pd.DataFrame) else DefaultSchema
+    def __init__(
+        self,
+        source: str | Path | list | pd.DataFrame | DataReader | Dataset,
+        schema: Schema | None = None,
+        *args,
+        **kwargs,
+    ) -> None:
+        # NOTE: this function currently assumes that DataReader will always be provided with a schema;
```
sivaprasadsudhir, line 127: I am not a big fan of how this is implemented now. Maybe, we can fix this in a later PR. But I think we should distinguish between a parent and a reader. 

A dataset should have a member reader. This constructor should take in an argument reader: str | Path | list | pd.DataFrame | DataReader and set the reader member. 

Constructor of the class Set should take an argument parent which should be of type Set | None. It should be set to None here (the first dataset in a query dag). 

This is cleaner and you can avoid some of the isinstance(node/source, Datareader) calls here and in your optimizer.

src/palimpzest/utils/data_helpers.py
```diff
@@ -0,0 +1,61 @@
...
+                path=path, pdfprocessor=pdfprocessor, file_cache_dir=file_cache_dir
+            )
+
+        elif all([f.endswith(tuple(constants.XLS_EXTENSIONS)) for f in os.listdir(path)]):
+            return XLSFileDirectoryReader(path)
+
+        elif all([f.endswith(tuple(constants.HTML_EXTENSIONS)) for f in os.listdir(path)]):
+            return HTMLFileDirectoryReader(path)
+
+        else:
```
sivaprasadsudhir, line 41: This is kinda bug-prone and could come back to bite us later.

I don't have a good solution though. Enforcing extensions for text files does not seem like a good idea.

sivaprasadsudhir (commented):
I did not review MAB/ random sentinel processor as Matt said they are in flux.

src/palimpzest/sets.py
```diff
@@ -129,52 +108,33 @@ def json_schema(self):
...
+        #       we will relax this assumption in a subsequent PR
+        # convert source into a DataReader
+        updated_source = get_datareader(source, **kwargs) if isinstance(source, (str, list, pd.DataFrame)) else source
```
chjuncn, line 130: How about Path?

src/palimpzest/utils/data_helpers.py
```diff
@@ -0,0 +1,61 @@
...
+    HTMLFileDirectoryReader,
+    ImageFileDirectoryReader,
+    MemoryReader,
+    PDFFileDirectoryReader,
+    TextFileDirectoryReader,
+    XLSFileDirectoryReader,
+)
+
+
+def get_local_source(path: str | Path, **kwargs) -> DataReader:
```
chjuncn, line 19: minor: 
1. rename this file to data_reader_helpers?
2. rename this function to get_local_data_reader?

src/palimpzest/query/operators/physical.py
```diff
@@ -168,16 +180,16 @@ def naive_cost_estimates(self, source_op_cost_estimates: OperatorCostEstimates)
...
+        correct estimates of this quantity).
         """
         raise NotImplementedError("CostEstimates from abstract method")
 
     def __call__(self, candidate: DataRecord) -> DataRecordSet:
         raise NotImplementedError("Calling __call__ from abstract method")
 
     @staticmethod
-    def execute_op_wrapper(operator: PhysicalOperator, op_input: DataRecord | list[DataRecord]) -> tuple[DataRecordSet, PhysicalOperator]:
+    def execute_op_wrapper(operator: PhysicalOperator, op_input: DataRecord | list[DataRecord] | int) -> tuple[DataRecordSet, PhysicalOperator]:
```
chjuncn, line 192: Do you mean dict? As the reader output is dict.

src/palimpzest/core/data/datareaders.py
```diff
@@ -0,0 +1,498 @@
...
+
+    def __str__(self) -> str:
+        return f"{self.__class__.__name__}(schema={self.schema})"
+
+    @property
+    def schema(self) -> Schema:
+        return self._schema
+
+    # NOTE: currently used by optimizer to compute node id for DataReaders
+    def serialize(self) -> dict:
```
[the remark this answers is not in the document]
mdr223: This exact iteration of it maybe, but the necessity of node ids / serializing nodes is a must have in order for the optimizer to cost the same node across multiple plans.

src/palimpzest/utils/data_helpers.py
mdr223: I agree -- if it's any consolation, the "`TextFileDirectoryReader`" calls `f.read()` on the each input file, so it should work on a pretty wide array of files.

But I agree that this approach is brittle and error prone -- I'd be happy to consider alternatives.

src/palimpzest/sets.py
mdr223: Good catch! Will fix that

src/palimpzest/utils/data_helpers.py
mdr223: Good idea, will make these changes

mdr223: (Going to keep `datareader` one word to match the filename(s) elsewhere, but happy to update it every where to `data_reader` if preferred)

PR #104, opened 2025-02-03
Refactor demos to use .sem_add_columns or .add_columns instead of convert(), remove Schema from demos when possible.

chjuncn:
This change is based on https://github.com/mitdbg/palimpzest/pull/101 and https://github.com/mitdbg/palimpzest/pull/102, please review them first then this change.

1. This is to refactor all demos to use .sem_add_columns or .add_columns, and remove .convert().

2. Remove Schema from demos, except demos using ValidationDataSource and dataset.retrieve() that need schema now. We can refactor these cases later.


src/palimpzest/sets.py
```diff
@@ -279,26 +284,28 @@ def add_columns(self, udf: Callable,
...
-        for col_name, col_type in types.items():
-            new_col = {}
-            new_col["type"] = col_type
-            new_col["desc"] = "New column: " + col_name
-            new_col["name"] = col_name
-            updated_cols.append(new_col)
+        for col_dict in types:
+            assert "name" in col_dict, "each type must contain a 'name' key specifying the column name"
+            assert "type" in col_dict, "each type must contain a 'type' key specifying the column type"
+            col_dict["desc"] = col_dict.get("desc", "New column: " + col_dict["name"])
```
sivaprasadsudhir, line 307: nit: why don't we make desc mandatory too?

src/palimpzest/core/data/datasources.py
```diff
@@ -41,8 +41,12 @@ class AbstractDataSource(abc.ABC):
     In the future, programmers can implement their own DataSources using custom Schemas.
     """
 
-    def __init__(self, schema: Schema) -> None:
-        self._schema = schema
+    def __init__(self, schema: type[Schema] | list[dict]) -> None:
```
sivaprasadsudhir, line 44: Should we continue to support Schema?

mdr223: I would like to, because I think some power users in the future may agree with our initial gut feeling that a class-based approach to defining fields & schemas was elegant, and there's little harm in supporting both

src/palimpzest/sets.py
mdr223: I think the main reason is because it can make some demos visibly shorter, and for some fields like `price` in the real estate example it's self-explanatory to the point of not being necessary.

I agree however that we should strongly encourage users to provide that field

mdr223 (approved):
I need to make one small change to allow for `Schema` to also be an input to `Dataset.sem_add_columns()` (in addition to the new column syntax), but once that's finished -- this is approved for merging!

PR #115, opened 2025-02-06
Update Docs with New Research Papers

mdr223:
Added new page which contains timeline of our research papers.

Added links on homepage to PalimpChat and Research Papers pages.


Tranway1 (approved):
Looks good to me

PR #118, opened 2025-02-10
Multi-LLM Refinement Pipeline for Query Output Validation

mdr223:
## Summary of PR

This PR contains the work to add a new `CriticConvert` physical operator to PZ. At a high-level, this operator runs a bonded convert, and then asks a critic model if the answer produced by the bonded convert can be improved upon. The original output and the critique are then fed into a refinement model, which produces the improved output.

The work to implement this includes:
1. Defining the physical operator in `src/palimpzest/query/operators/critique_and_refine_convert.py`
2. Adding an implementation rule for this physical operator in `src/palimpzest/query/optimizer/rules.py`
3. Adding boolean flag(s) to enable allowing / disallowing this physical optimization
4. Adding base prompts for the critique and refinement generations

One other change which this work spawned was an attempt to improve the management and construction of our prompts -- and to decouple this logic from the `BaseGenerator` class. On the management side, I split our single `prompts.py` file into a set of files. On the construction side, I created a `PromptFactory` class which templates prompts based on the `prompt_strategy` and input record.

The `PromptFactory` is not a perfect solution, but I think it is a step in the right direction.

Finally, I fixed an error which previously filtered out `RAGConvert` operators from being considered by the `Optimizer`, and I made 2-3 more miscellaneous small tweaks.


src/palimpzest/prompts/prompt_factory.py
```diff
@@ -0,0 +1,721 @@
...
+            while total_context_len * TOKENS_PER_CHARACTER > 6000:
+                # sort fields by length
+                field_lengths = [(field, len(value)) for field, value in context.items()]
+                sorted_fields = sorted(field_lengths, key=lambda item: item[1], reverse=True)
+
+                # get field with longest context
+                longest_field_name, longest_field_length = sorted_fields[0]
+
+                # trim the field
+                context_factor =  6000.0 / (total_context_len * TOKENS_PER_CHARACTER)
```
chjuncn, line 140: Can you add a todo for chunking so that we don't need to do trim here? 

I feel silently trimming information is not ideal, if we can add information somewhere saying we're trimming would be great before we do chunking. But if it's not easy to do, that's fine.

src/palimpzest/query/operators/critique_and_refine_convert.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def __str__(self):
+        op = super().__str__()
+        op += f"    Critic Model: {self.critic_model}\n"
+        op += f"    Critic Prompt Strategy: {self.critic_prompt_strategy}\n"
+        op += f"    Refine Model: {self.refine_model}\n"
+        op += f"    Refinement Prompt Strategy: {self.refinement_prompt_strategy}\n"
+        return op
+
+    def get_id_params(self):
```
chjuncn, line 49: This is not really related to this PR, but I feel get_id_params is confusing to me. Why not just provide get_id() or id() or or universal_id()? 

Generally, I see we have ids in different places, universal_id, id, and so on, and think they're not intuitive to understand. Maybe we should revisit these id generators or naming sometime, creating an issue should be good enough.

src/palimpzest/query/optimizer/rules.py
```diff
@@ -649,6 +650,97 @@ def substitute(cls, logical_expression: LogicalExpression, **physical_op_params)
...
+            # 1. this is a pure vision model and we're not doing an image conversion, or
+            # 2. this is a pure text model and we're doing an image conversion, or
+            # 3. this is a vision model hosted by Together (i.e. LLAMA3_V) and there is more than one image field
+            first_criteria = model in pure_vision_models and not is_image_conversion
+            second_criteria = model in pure_text_models and is_image_conversion
+            third_criteria = model == Model.LLAMA3_V and (num_image_fields > 1 or list_image_field)
+            if first_criteria or second_criteria or third_criteria:
+                continue
+
+            models.append(model)
```
chjuncn, line 716: Probably add a todo: centralize model selection or filtering in a new class or util, we don't need to expose such detail in this level of class.

This comment is from line681.

src/palimpzest/query/optimizer/rules.py
```diff
@@ -649,6 +650,97 @@ def substitute(cls, logical_expression: LogicalExpression, **physical_op_params)
 
         return set(physical_expressions)
 
+class CriticConvertRule(ImplementationRule):
```
chjuncn, line 653: Add a todo: Many ConverRule share same set of code, consider to reduce code duplication.

src/palimpzest/query/optimizer/rules.py
```diff
@@ -649,6 +650,97 @@ def substitute(cls, logical_expression: LogicalExpression, **physical_op_params)
...
+            if first_criteria or second_criteria or third_criteria:
+                continue
+
+            models.append(model)
+
+        # construct CriticConvert operations for every combination of model, critic model, and refinement model
+        physical_expressions = []
+        for model in models:
+            for critic_model in models:
+                for refine_model in models:
```
chjuncn, line 722: Related to previous comment: 

The current candidate generation is brute-force solution. From the latest research we can see some models are better to be refine, and others are better for critic, so I'm imagining ModelSelection could be a new class, we will have different model selections for different cases. For example, different convertRule will have different model selection strategy, and this kind of detail could be in happen in a function instead of spreading here. 

Adding a todo is good enough for now.

chjuncn:
The prompt_factory design definitely is better than before, the generator is clearer than before. Thank you! I think it's good enough for now, if we see more advanced requirements for prompts in the future, we could extract more things from the factory, but now we're good. 

I give some design comments on the files I see, and adding todos or creating new issues is good enough, I don't expect we resolve that in this PR.

# Open Question
One question for us to think for adding new Physicalop: For example, when we add a new converter physicalOp, can we add unit tests for: 1. this converter will be sure selected or not selected in some cases? 2. This converter is using the expected models and reduce the right results? 

Adding new unit test for new rule or new Physicalop should be considered, but I don't want to block this PR by this, so adding unit tests should be following PRs.

# What I didn't do in this review:
1. didn't check details for each *prompts.py
2. didn't run tests to make sure there is no regression
3. didn't make sure if the `CriticConvert`  works as expected.  If you have done the experiments for this converter, can you report metrics like: when this converter will be selected usually? how long users get final output in general? or other information that you think is useful for users to know this converter in general. Adding these metrics as comments in `critique_and_refine_convert.py` is good enough.

If you expect me to do these tests, please let me know.

src/palimpzest/prompts/prompt_factory.py
```diff
@@ -0,0 +1,721 @@
...
+
+        # TODO: this does not work for image prompts
+        # TODO: this ignores the size of the `orignal_messages` in critique and refine prompts
+        # cut down on context based on window length
+        if self.model in [Model.LLAMA3, Model.MIXTRAL]:
+            total_context_len = len(json.dumps(context, indent=2))
+
+            # sort fields by length and progressively strip from the longest field until it is short enough;
+            # NOTE: 6000 is a rough estimate which leaves room for the rest of the prompt text
+            while total_context_len * TOKENS_PER_CHARACTER > 6000:
```
sivaprasadsudhir, line 131: nit: Can you make this 6000 a constant (or at least a variable)?

src/palimpzest/query/operators/critique_and_refine_convert.py
```diff
@@ -0,0 +1,112 @@
...
+from palimpzest.core.data.dataclasses import GenerationStats, OperatorCostEstimates
+from palimpzest.core.elements.records import DataRecord
+from palimpzest.query.generators.generators import generator_factory
+from palimpzest.query.operators.convert import LLMConvert
+
+# TYPE DEFINITIONS
+FieldName = str
+
+
+class CriticConvert(LLMConvert):
```
sivaprasadsudhir, line 15: nit: can we rename this to CriticAndRefineConvert?

sivaprasadsudhir (approved):

src/palimpzest/query/operators/critique_and_refine_convert.py
mdr223: Great question! The `get_id_params()`, `get_op_params()`, and `universal_identifier()` are not well documented, but they all play an important (and distinct) role in PZ. I'd like to take this opportunity to expand on their purpose(s) as they relate to the Optimizer, and maybe as a group we can come up with a better approach that is slightly less opaque.

At a very high-level, the Optimizer's job is to find the best physical plan for a user program. To do this, we follow a recipe that looks something like the following:

1. Convert the user program to a (set of) logical plan(s)
2. Convert those logical plan(s) into a set of physical plans
3. Cost the physical plans and return the best one

While this sounds simple, there are a lot of implementation details that you need to get right. One such detail is determining whether two physical operators across two different physical plans are actually the same operator. 

Why is this important? Currently in PZ, our cost model is built around sampled data. As the number of physical operators grows, it becomes expensive to sample every physical operator (let alone every physical plan) to get high quality cost, runtime, and quality information. Thus, in an effort to be economical with our samples, we need to re-use the cost information gathered about a physical operator in one plan and apply it to that operator in other plans. (This of course rests on an assumption that operator performance is independent of its placement within a plan -- which is not always true -- but for now this is an assumption I am willing to make).

To illustrate this^ point, consider the following two plans:

Plan 1: scan --> convert A1 --> filter B1
Plan 2: scan --> convert A2 --> filter B2
Plan 3: scan --> filter B1 --> convert A1
Plan 4: scan --> filter B2 --> convert A1

Let's say that we sample Plans 1 and 2 by processing a few inputs with each plan. Now, without having sampled Plans 3 and 4, we want to produce a cost estimate for them. Let's assume our cost model for Plans 3 and 4 is:
```
plan_dollar_cost = sum([op.cost for op in plan])
plan_runtime = sum([op.runtime for op in plan])
plan_quality = product([op.quality for op in plan])
```
(I realize that this^ cost model is not perfect, but let's just assume it is our cost model of choice for now.)

When we sample Plans 1 and 2, we compute the average cost, runtime, and quality of each operator and use this as our estimate for that operator. We store this information in a dictionary:
```
operator_to_stats = {
  "convert_A1_physical_op_id": {"cost": 1.0, "runtime": 2.0, "quality": 0.5, "selectivity": 1.0},
  "convert_A2_physical_op_id": {"cost": 1.5", "runtime": 4.0, "quality": 0.9, "selectivity": 1.0},
  "filter_B1_physical_op_id": {"cost": 0.75, "runtime": 0.5, "quality": 0.8, "selectivity": 0.3},
  "filter_B2_physical_op_id": {"cost": 2.75, "runtime": 0.7, "quality": 0.9, "selectivity": 0.4},
}
```
Now, to estimate Plan 3 using our cost model above, we simply need to compute:

```
plan3_dollar_cost = sum([operator_to_stats[op.get_op_id()]["cost"] for op in plan3])
plan3_runtime = sum([operator_to_stats[op.get_op_id()]["runtime"] for op in plan3])
plan3_quality = product([operator_to_stats[op.get_op_id()]["quality"] for op in plan3])
```

**Here's the key issue: in order for `operator_to_stats[op.get_op_id()]` to lookup the correct entry in the dictionary, it must be the case that `op.get_op_id()` is:**

1. **Independent of the operator's location in the plan**
2. **Dependent on the setting of the operator's parameters**

Criteria (1.) implies that `op.get_op_id()` cannot be a function of:

1. the `op_id` of the parent operator
2. the `input_schema` of the operator, and
3. the `output_schema` of the operator, as `sem_add_columns()` now computes `output_schema = self.schema.union(output_schema)` (i.e. it also contains the input schema's fields in order to accurately reflect the schema of a `DataRecord` after being processed by that operator).

Criteria (2.) implies that we still need to specify which of an operator's parameters make that operator **unique** in the eyes of our optimization. For example, the `target_cache_id` cannot make a physical operator unique, because it is a function of the parent operator's `target_cache_id` (which violates criteria (1.)).

As a result of this^, the approach I settled on was to have each subclass of `PhysicalOperator` define the set of parameters make that operator **unique**. These should not include things like `input_schema`, `output_schema`, and `target_cache_id`, but they should include things like `model`, `filter`, `agg_func`, etc. **This is the purpose that `get_id_params()` serves. Its output is a dictionary containing these unique key-value pairs for each physical operator.**

On the one hand, this approach places a burden on the PZ developer because it requires them to implement this dictionary for each new physical operator that they create. One might be tempted to instead enumerate the set of physical operator parameters which should NOT be included (e.g. `input_schema`, `output_schema`, `target_cache_id`, etc.) and compute an operator's `get_id_params()` as the set of all `kwargs` which are not in this disallowed list. However, with this^ approach we still need someone (the developer or PR reviewer) to check whether a parameter of the new physical operator needs to be added to the disallow list.

Personally, I opted for the `get_id_params()` approach (i.e. having the developer explicitly state the unique parameters), because I figured that -- at a minimum -- this would force a new PZ developer to stop and ask "Hey, what should this function return?". In the other approach, I think it's too easy for a new parameter to sneak into the set of unique params without first being added to a disallow list.

Finally, this begs the question:

1. what does `get_op_id()` do?
2. what does `get_logical_op_id()` (and `get_logical_id_params()`) do?
3. what does `get_op_params()` / `get_logical_op_params()` do?
4. and what does `universal_identifier()` / `get_node_uid()` do?

The answers are:

1. `get_op_id()` calls `get_id_params()` and converts the output into a string
2. These are the logical operator equivalents of `get_op_id()` and `get_id_params()`
3. `get_op_params()` / `get_logical_op_params()` returns **the full set of physical / logical** operator parameters for when we need to make a copy of an operator. This ***should*** include parameters like `input_schema`, `output_schema`, etc.
4. There are some cases where we do want a **universally unique identifier for a Dataset / logical operator** (e.g. when computing the `target_cache_id`, or when disambiguating between two identical field names in two different output schemas. `universal_identifier()` computes a hash which includes the parent's `universal_identifier()` (and all `Dataset` parameters), thus this can be used for both `target_cache_id` and field disambiguation.

I hope this^ provides more context on some of these opaque functions within PZ. I do not believe that I have found a perfect implementation, but with this context in mind, maybe we can all come up with a cleaner implementation of these stated goals.

FYI: @sivaprasadsudhir @mikecafarella @vitaglianog @Tranway1

mdr223: Created #126

src/palimpzest/query/optimizer/rules.py
mdr223: Agreed that the logic to find the appropriate (set of) model(s) for many of these rules can be moved into one or more utility functions.

Created #127

src/palimpzest/query/optimizer/rules.py
mdr223: See #127

src/palimpzest/query/optimizer/rules.py
mdr223: I think that what you describe is a good idea: using prior knowledge to limit the subset of models that we consider for various operations (e.g. critiquing or refining).

That being said, at the moment I would like to keep this logic close to each individual rule, rather than abstracting it away from the rules.

My motivation for this viewpoint comes from a talk about the Cascades optimizer in Microsoft SQL Server. In their system, the optimizer needs to consider a space of plans that is orders of magnitude larger than the space of plans we have now. Furthermore, their optimizer needs to return an optimal plan in a much shorter timeframe than we are allowed to (think milliseconds).

How do they accomplish this? The answer is a combination of:

1. aggressive pruning logic (to eliminate entire subtrees in the search), and
2. a lot of heuristics

What you describe (limiting the set of models we consider for critiquing or refining) sounds to me like a good heuristic. In practice, many of our implementation rules will evolve to have heuristics like this one in order to shrink the space of the search.

One challenge this presented in SQL Server was making sure that new heuristics (and new rules) did not create regressions for large sets of queries. What they ultimately did was create a test suite of queries which they ran against the Optimizer. If they added a new rule (or heuristic) and saw the performance regress, then they could investigate the bug.

However, even with a very large test bed, you cannot have full coverage over all possible queries. My fear is that if we abstract the logic for `ModelSelection` into a separate class -- which is shared across all Rules -- and then we update it when we add some new Rule, it is possible that this addition can silently degrade the performance of the existing rules which users have come to rely on.

By tying logic like "what is the set of critic models?" to the `CriticConvertRule` itself, we can more easily isolate any regressions to the new presence of the `CriticConvertRule`, rather than to a change in e.g. `ModelSelection` logic which affects all the Rules which use its logic.

An easy way to do this is to maintain a class attribute at the top of `CriticConvertRule` which enumerates the models that can be use for e.g. critiquing or refining. An example of this is in our `TokenReducedConvertRule`, where our heuristic is that we only need to consider `token_budgets = [0.1, 0.5, 0.9]`.

Thus, my preferred approach to implementing what you describe would be to have something like:
```
class CriticConvertRule(ImplementationRule):
    """
    Implementation rule for the CriticConvert operator.
    """
    critic_models = [...]
    refine_models = [...]

    ...

    @classmethod
    def substitute(cls, ...) -> ...:
        ...
        models = # logic for selecting the initial model

        # construct CriticConvert operations for every combination of model, critic model, and refinement model
        physical_expressions = []
        for model in models:
            for critic_model in cls.critic_models:
                for refine_model in cls.refine_models:
                    ...

        return set(physical_expressions)
```
In addition to the motivations described above, this also makes it easy for other developers to figure out what heuristics have gone into the search space (i.e. they can just read the class attributes at the top of the class).

src/palimpzest/prompts/prompt_factory.py
mdr223: Sure thing!

src/palimpzest/query/operators/critique_and_refine_convert.py
mdr223: Agreed, that is cleaner

mdr223:
I think the following is a very good design goal:
> ## Open Question
> One question for us to think for adding new Physicalop: For example, when we add a new converter physicalOp, can we add unit tests for: 1. this converter will be sure selected or not selected in some cases? 2. This converter is using the expected models and reduce the right results?
I currently have some tests which check for (1.) in some cases -- but it is nowhere near full coverage of our rules. This will require a significant effort to curate test fixtures (and data) which we can run our tests against -- but it is 100% a price worth paying IMO.

sivaprasadsudhir (approved):
Looks good to me.

PR #128, opened 2025-02-17
first commit

lilbarbar:


mdr223:
Hi @lilbarbar and @7ama2004,

One quick comment I have is that -- generally speaking -- you never want to push virtual environments to remote Github repositories. It's not a big deal, it's just that they contain a lot of files and are relatively large in size.

Instead, we typically push files like `pyproject.toml` (or `requirements.txt` in older python projects) which contain all of the dependencies that one might need.

I would encourage you to make a second commit in which you:
- remove the `myenv` directory from the git repo (see below for instructions)
- add `myenv/` to the `.gitignore` to prevent yourself from accidentally pushing your virtualenv in the future

For the first bullet: you want to remove `myenv` from the git repository, ***but you don't want to delete it entirely from your filesystem.*** (If you delete it entirely from your filesystem, then your virtual environment will be gone!)

To remove `myenv` from Git *without deleting it from your local filesystem* take a look at the answer to this StackOverflow post: https://stackoverflow.com/a/1143800

Let me know if you have any questions!

demos/biofabric-demo.py
```diff
@@ -87,7 +87,7 @@ def print_table(output):
     parser.add_argument("--executor", type=str, help="The plan executor to use. The avaliable executors are: sequential, pipelined_parallel, pipelined_single_thread", default="pipelined_parallel")
 
     args = parser.parse_args()
-    no_cache = args.no_cache
+    cache = args.cache
```
mdr223, line 90: self TODO: remove

demos/image-demo.py
```diff
@@ -32,10 +32,10 @@ def build_image_plan(dataset):
     start_time = time.time()
 
     parser = argparse.ArgumentParser(description="Run a simple demo")
-    parser.add_argument("--no-cache", action="store_true", help="Do not use cached results", default=True)
+    parser.add_argument("cache", action="store_false", help="Use cached results", default=False)
```
mdr223, line 35: TODO: fix this to have `--cache` and keep `store_true`

mdr223: For more detail, `action="store_true"` makes the argument parser store a default value of `False` for the variable if the argument is not provided by the user (the name is confusing). Thus, we are somewhat redundantly specifying `action="store_true"` AND `default=False`.

But I am happy to have this redundancy as it makes the code a bit more explicit and clear, so if you make the changes specified in the previous comment that should be good.

More info: https://stackoverflow.com/questions/8203622/argparse-store-false-if-unspecified#:~:text=The%20store_true%20option%20automatically%20creates,line%20argument%20is%20not%20present.

src/palimpzest/sets.py
```diff
@@ -42,7 +42,7 @@ def __init__(
         limit: int | None = None,
         cardinality: Cardinality = Cardinality.ONE_TO_ONE,
         depends_on: list[str] | None = None,
-        nocache: bool = False,
+        cache: bool = True,
```
mdr223, line 45: TODO: It looks like we previously had a bug here because we specified `nocache = False` by default. You correctly flipped the logic, but what I think we actually want is: `cache: bool = False`

src/palimpzest/sets.py
```diff
@@ -60,7 +60,7 @@ def __init__(
         self._limit = limit
         self._cardinality = cardinality
         self._depends_on = [] if depends_on is None else sorted(depends_on)
-        self._nocache = nocache
+        self.cache = cache
```
mdr223, line 63: TODO: this should be `self._cache = cache` (the class member should be prefixed with an underscore to keep convention with the other class members).

src/palimpzest/sets.py
```diff
@@ -181,7 +181,7 @@ def sem_filter(
             schema=self.schema,
             filter=f,
             depends_on=depends_on,
-            nocache=self._nocache,
+            cache=self.cache,
```
mdr223, line 184: TODO: `cache=self._cache`

src/palimpzest/query/optimizer/optimizer.py
```diff
@@ -86,7 +86,7 @@ def __init__(
         self,
         policy: Policy,
         cost_model: CostModel,
-        no_cache: bool = False,
+        cache: bool = True,
```
mdr223, line 89: TODO: looks like this was another error that existed from before, we should set this to `cache: bool = False`

mdr223 (changes requested):
Hi @lilbarbar and @7ama2004, I've left TODO comments in a few places where I found some minor fixes that need to happen. Once you fix those and commit the changes, I can take a look again and hopefully merge into dev!

7ama2004:
> Hi @lilbarbar and @7ama2004, I've left TODO comments in a few places where I found some minor fixes that need to happen. Once you fix those and commit the changes, I can take a look again and hopefully merge into dev!

7ama2004:
> > Hi @lilbarbar and @7ama2004, I've left TODO comments in a few places where I found some minor fixes that need to happen. Once you fix those and commit the changes, I can take a look again and hopefully merge into dev!

Hi Matt, I implemented the changes, just to make sure,

mdr223 (approved):
Thanks for making those fixes, this looks good to go!

PR #132, opened 2025-02-19
Fix the MoA prompt generation

sivaprasadsudhir:
Set `output_fields_desc` for MoA proposer and aggregator prompts.


mdr223 (approved):

mdr223:
@sivaprasadsudhir my only comment would be that this should go into `main` as a hotfix

PR #134, opened 2025-02-20
Add logger for PZ

chjuncn:
add logger for PZ
1. When verbose=True, we save all logs to log_file and print them on  console; This can be adjusted later.
2. when verbose=False, we only save ERROR+ log to file and print ERROR+.

I just add logging to somewhere I think might be important for the execution, we always can add/remove for more or less.

Also I might update the logging message based on my later annotation work. But this PR should setup the logging mechanism for now.

## Future improvements
1. We can consider use verbose_level instead of true/false to control logging
2. When we consider to scale PZ, logging service is necessary so that it won't block the program.


src/palimpzest/query/optimizer/rules.py
```diff
@@ -237,6 +254,8 @@ def substitute(cls, logical_expression: LogicalExpression, **physical_op_params)
             generated_fields=logical_expression.generated_fields,
             group_id=logical_expression.group_id,
         )
```
mdr223, line 256: should we also do the `deduped_physical_expressions` pattern that is used below here?

src/palimpzest/query/optimizer/rules.py
```diff
@@ -325,6 +349,8 @@ def substitute(cls, logical_expression: LogicalExpression, **physical_op_params)
             )
             physical_expressions.append(expression)
```
mdr223, line 351: should we also do the `deduped_physical_expressions` pattern that is used below here?

src/palimpzest/query/processor/nosentinel_processor.py
```diff
@@ -215,6 +223,9 @@ def execute_plan(self, plan: PhysicalPlan, num_samples: int | float = float("inf
             if self.progress_manager:
                 self.progress_manager.finish()
 
+        logger.info(f"Done executing plan: {plan}")
+        logger.debug(f"Output records: {output_records}")
+        logger.debug(f"Plan stats: {plan_stats}")
```
mdr223, line 228: Same comment here as before re: selectively logging `plan_stats`; I also think you might have a similar issue with `output_records` if the list is very large.

src/palimpzest/query/processor/nosentinel_processor.py
```diff
@@ -404,6 +418,9 @@ def execute_plan(self, plan: PhysicalPlan, num_samples: int | float = float("inf
...
```
mdr223, line 423: Same comment here as above re: `plan_stats` and `output_records`

src/palimpzest/query/processor/query_processor.py
```diff
@@ -172,6 +178,10 @@ def execute_plans(
             if plan.plan_id == max_quality_plan_id:
                 return_records = records
 
+        logger.info(f"Done executing plans: {plans}")
+        logger.debug(f"All sample execution data: {all_sample_execution_data}")
+        logger.debug(f"Return records: {return_records}")
+        logger.debug(f"All plan stats: {all_plan_stats}")
```
mdr223, line 184: Same comment here as before re:
- `all_sample_execution_data`
- `return_records`
- `all_plan_stats`

mdr223 (changes requested):
Overall, this is a great addition to PZ!

I left a few comments, some of them are minor nits, but some I think should be addressed in instances where we may spam the logs with massive data dumps.

src/palimpzest/tools/logger.py
```diff
@@ -0,0 +1,101 @@
...
+            "timestamp": self.formatTime(record),
+            "level": record.levelname,
+            "logger": record.name,
+            "message": record.getMessage(),
+        }
+
+        # Add extra fields if they exist (for metrics/stats)
+        if hasattr(record, "stats"):
+            log_data["stats"] = record.stats
+        if record.exc_info:
```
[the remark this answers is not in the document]
chjuncn: good catch! fixed.

src/palimpzest/query/execution/parallel_execution_strategy.py
```diff
@@ -207,8 +208,10 @@ def execute_plan(self, plan: PhysicalPlan, num_samples: int | float = float("inf
                 # self.datadir.close_cache(operator.target_cache_id)
                 pass
 
+        logger.info(f"Completed execution of plan {plan.plan_id} in {time.time() - plan_start_time:.2f} seconds")
         # finalize plan stats
         total_plan_time = time.time() - plan_start_time
         plan_stats.finalize(total_plan_time)
+        logger.debug(f"Plan execution stats: {plan_stats}")
```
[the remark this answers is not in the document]
chjuncn: plan_stats maybe is too much. I'm not 100% sure what information I need for now, I'll use your suggestion and adjust later if I need more info for the annotation tool. :)

src/palimpzest/query/execution/single_threaded_execution_strategy.py
```diff
@@ -129,6 +134,9 @@ def execute_plan(self, plan: PhysicalPlan, num_samples: int | float = float("inf
         total_plan_time = time.time() - plan_start_time
         plan_stats.finalize(total_plan_time)
 
+        logger.info(f"Completed execution of plan {plan.plan_id} in {time.time() - plan_start_time:.2f} seconds")
+        logger.debug(f"Plan execution stats: {plan_stats}")
```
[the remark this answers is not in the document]
chjuncn: done

src/palimpzest/query/processor/query_processor.py
chjuncn: I change them to numbers, and will adjust in following PRs based on the need.

PR #138, opened 2025-02-20
Begin Process of Improving Index Abstraction(s) in PZ

mdr223:
This PR contains a very rough initial cut of work to support cost tracking for indices / retrieve operations in PZ.

@sivaprasadsudhir and I are working on a more robust abstraction which we will contribute in a separate PR.

For now, this PR enables the user to provide a `RAGPretrainedModel` or `chromadb` (with `text-embedding-3-small`) index  which we can compute query embeddings for (in the case of `chromadb`). This enables us to track the cost of computing these embeddings.

### DIFF
--- a/src/palimpzest/query/operators/retrieve.py
+++ b/src/palimpzest/query/operators/retrieve.py
@@ -242,6 +242,7 @@ class RetrieveOp(PhysicalOperator):
             with open(f"retrieve-errors/error-{ts}.txt", "w") as f:
                 f.write(str(query))
 
+        # TODO: the user is always right! let's drop this post-processing in the future
         # filter top_results for the top_k_results
         top_k_results = []
         if all([isinstance(result, list) for result in top_results]):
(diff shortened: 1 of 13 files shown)


src/palimpzest/query/operators/retrieve.py
```diff
@@ -92,19 +164,96 @@ def __call__(self, candidate: DataRecord) -> DataRecordSet:
...
+            top_results = self.search_func(self.index, query, self.k)
+        except Exception:
+            top_results = ["error-in-retrieve"]
+            os.makedirs("retrieve-errors", exist_ok=True)
+            ts = time.time()
+            with open(f"retrieve-errors/error-{ts}.txt", "w") as f:
+                f.write(str(query))
+
+        # filter top_results for the top_k_results
+        top_k_results = []
```
sivaprasadsudhir, line 246: Is this for the case the user-provided search_func does not return <= k entries? Should we be overriding the user in that case?

sivaprasadsudhir (approved):

PR #140, opened 2025-02-21
Use a singleton API client for operators

sivaprasadsudhir:
Share the API client instance across all operators by creating a singleton. This is the intended way to use these clients. Also, this is needed as these API clients (at least openAI client) cannot be closed and we eventually run out of file descriptors with too many open connections. 

Currently all operators that use a Generator will use the shared instance. Retrieve and RAGConverts will be fixed in a later PR.


mdr223 (approved):
Fantastique

PR #146, opened 2025-02-22
use NullHandler in __init__ and let application control logging config

chjuncn:


sivaprasadsudhir (approved):
Overall, this looks good to me. 

My only question is whether we need a different logger for each file. User will need to know the name of each logger to set its log levels. Maybe, one logger per submodule is sufficient? We can iterate on this based on our experience of using the loggers.

chjuncn:
Thanks! I think it'd be useful to have file indication which will be easy to debug. And I agree let's update based on our learnings moving forward.

PR #149, opened 2025-02-25
Eliminate confidence interval stuff

mdr223:
## Summary of PR
We never ended up using the confidence interval approach to optimization as we quickly pivoted to MAB. The code has survived for a while but its time has come to an end.

### DIFF
--- a/src/palimpzest/query/optimizer/cost_model.py
+++ b/src/palimpzest/query/optimizer/cost_model.py
@@ -13,7 +13,6 @@ import warnings
 from typing import Any
 
 import pandas as pd
-import scipy.stats as stats
 
 from palimpzest.constants import MODEL_CARDS, NAIVE_BYTES_PER_RECORD, GPT_4o_MODEL_CARD, Model
 from palimpzest.core.data.dataclasses import OperatorCostEstimates, PlanCost, RecordOpStats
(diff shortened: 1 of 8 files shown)


src/palimpzest/query/optimizer/cost_model.py
```diff
@@ -548,64 +468,36 @@ def _compute_operator_estimates(self) -> dict[str, Any] | None:
...
-                    "time_per_record": time_per_record,
-                    "time_per_record_lower_bound": time_per_record_lb,
-                    "time_per_record_upper_bound": time_per_record_ub,
-                    "selectivity": selectivity,
-                    "selectivity_lower_bound": selectivity_lb,
-                    "selectivity_upper_bound": selectivity_ub,
-                }
+                time_per_record = self._est_time_per_record(op_df)
+                selectivity = self._est_selectivity(self.sample_execution_data_df, op_df)
+                estimates = {"time_per_record": time_per_record, "selectivity": selectivity}
```
sivaprasadsudhir, line 491: I find selectivity confusing. Maybe, we should just call it fan_out?

sivaprasadsudhir (approved):

sivaprasadsudhir (approved):

src/palimpzest/query/optimizer/cost_model.py
mdr223: I agree, let's make this change universally in a separate PR once the other big-ticket items are merged.

mdr223: Created #151 for this

PR #158, opened 2025-02-27
Add Pneuma-Palimpzest Integration Demo

luthfibalaka:
# Pneuma-Palimpzest Integration Demo

Pneuma is an LLM-based data discovery system for tabular data. Given a natural-language question, Pneuma retrieves the most relevant table(s) for the question. However, in some cases, users may want direct answers instead of just the raw data. This is where Palimpzest (PZ) comes in. It builds on Pneuma's retrieval step by answering questions using the relevant tables Pneuma provides.

As an illustration, suppose a chemist has research questions about molecular properties. Instead of manually searching through large datasets, they need a system that can both retrieve relevant tables and extract insights from them. This demo showcases how Pneuma and PZ work together in such a scenario.

For this demo, we use a subset of the [ChEMBL dataset](https://www.ebi.ac.uk/chembl/), which is a database of bioactive molecules with drug-like properties. The questions are directly related to the dataset’s contents, such as _“What fifth-level classification is associated with molecules that have the registration number 1609148?”_ Pneuma first indexes ChEMBL using `pneuma-gen-index.py`, which also handles downloading the dataset and the associated questions (`questions.jsonl`). Then, in `pneuma-demo.py`, Pneuma retrieves relevant tables, and PZ generates answers based on them. This demonstrates the end-to-end pipeline, where Pneuma discovers relevant tables, and PZ transforms retrieved tables into direct, actionable insights.

### DIFF
--- a/demos/pneuma-demo.py
+++ b/demos/pneuma-demo.py
@@ -79,9 +79,6 @@ def parse_arguments():
 
 def build_pneuma_query(pneuma: Pneuma, dataset: QuestionDataReader, k: int):
     questions = pz.Dataset(dataset)
-    questions = questions.sem_add_columns(
-        question_dataset_cols, desc="Extract the question"
-    )
 
     def extract_table_content(df: pd.DataFrame):
         columns = " | ".join(df.columns)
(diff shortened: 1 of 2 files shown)


mdr223 (approved):
LGTM

PR #161, opened 2025-03-05
ArchonConvert Operator (generators -> critic -> fuser)

yash94404:
- Created new ArchonConvert operator that runs a convert prompt through a layer of multiple generators, which have all their outputs critiqued in the critic layer, and fused together into an optimal response in the fuser layer
- Implemented ArchonConvert class, new prompt strategies for Archon operations, new prompts for Archon
- Integrated into optimizer with ArchonConvertRule, allowing the optimizer to use any subset of models for the generator layer (can modify LLM architecture if needed)

Next Steps:
- Modifying LLM Architecture to support multiple critics/fusers, adding verifier/ranker?
- Correspondingly to above, improving optimizer using DAG approach talked about with Siva
- Any other ideas!


mdr223:
Closing temporarily; I want to revisit this PR in the near-term as part of a targeted curation of the physical operators implemented by PZ.

PR #162, opened 2025-03-05
Adding Support for Annotator Tool and ValidationData Class

mdr223:
WIP


demos/annotation.py
```diff
@@ -0,0 +1,43 @@
...
+]
+
+# lazily construct the computation to get emails about holidays sent in July
+#dataset = Dataset("testdata/matsci_pdfs/")
+dataset = Dataset("testdata/Sample_Papers_NLP_tiny")
+dataset = dataset.sem_add_columns(paper_cols)
+
+# execute the computation w/the MaxQuality policy
+config = pz.QueryProcessorConfig(verbose=True, execution_strategy="parallel")
+val_data = ValidationData(".pz_logs/execution_id_42e8004490/annotated_real.json")
```
mdr223, line 28: I'm happy to check in this demo file, but perhaps we should make this json part of `testdata/`?

demos/annotation.py
```diff
@@ -0,0 +1,43 @@
...
+val_data.set_score_fn("TargetMaterial", bleu_scorer)
+val_data.set_score_fn("MainComponents", bleu_scorer)
+val_data.set_score_fn("Strength", bleu_scorer)
+config.val_data = val_data
+
+config.processing_strategy = "mab_sentinel"
```
mdr223, line 34: Note to myself: I will need to update this once my PR is merged into `dev`

src/palimpzest/core/data/datareaders.py
```diff
@@ -443,11 +443,10 @@ def __getitem__(self, idx: int) -> dict:
                 }
         """
         filepath = self.filepaths[idx]
-        filename = os.path.basename(filepath)
         with open(filepath) as f:
             contents = f.read()
 
-        return {"filename": filename, "contents": contents}
+        return {"filename": filepath, "contents": contents}
```
mdr223, line 449: Out of curiosity, is there a specific reason for using `filepath` instead of `filename` (i.e. did `filename` cause an error somewhere?) We should also update other DataReaders to use `filepath` as well to make this consistent (e.g. `XLSFileDirectoryReader`) and perhaps also modify the `TextFile`, `PDFFile`, etc. schemas to reflect the change.

src/palimpzest/core/data/validationdata.py
```diff
@@ -0,0 +1,172 @@
...
+        # Read execution stats JSON
+        if self.file_path is not None:
+            with open(self.file_path) as f:
+                execution_stats = json.load(f)
+        elif self.annotations is not None:
+            execution_stats = self.annotations
+        else:
+            execution_stats = ExecutionStats().to_json()
+
+        if "plan_stats" in execution_stats:
```
mdr223, line 29: Note to self: now that `execution_stats` has a separate field for `sentinel_plan_stats` we may want to process that as well?

src/palimpzest/core/data/validationdata.py
```diff
@@ -0,0 +1,172 @@
...
+            working_stats = execution_stats["plan_stats"]
+            if len(working_stats) > 1:
+                _, working_stats = working_stats.popitem()
```
mdr223, line 32: or perhaps not, if this logic was trying to isolate the non-sentinel plan stats?

src/palimpzest/core/data/datareaders.py
mdr223: Ah, I see now that this was done to enable getting the input dataset in the `ValidationData` class. Let's perhaps revert this^ to being `filename` and see if we can come up with a more elegant way to provide the input dataset in `ValidationData`.

src/palimpzest/core/data/validationdata.py
```diff
@@ -0,0 +1,172 @@
...
+            logical_op_id = None
+            # Process each record's stats
+            for record_stats in op_stats["record_op_stats_lst"]:
+                logical_op_id = record_stats["logical_op_id"]
+                record_source_idx = record_stats["record_source_idx"]
+
+                # If there is no annotations, then we use the record_state and passed_operator
+                if "annotations" in record_stats:
+                    labels = record_stats["annotations"]["labels"]
+                    pass_filter = record_stats["annotations"]["pass_filter"]
```
mdr223, line 50: Let's use `passed_operator` instead of `pass_filter` here and below to match the term in `RecordOpStats` (and throughout the codebase).

(The reason for the more general term is that it makes it a little cleaner to compute operator selectivity / fan out in the cost model & optimizer code because I can assume every operator's `RecordOpStats` has a `passed_operator` field. I thought it would be a bit weird to have `passed_filter` as a field for non-filter operators, and I didn't like having an `if-else` based on the type of the operator).

src/palimpzest/query/processor/mab_sentinel_processor.py
mdr223, line 1: Will need to merge these changes w/my pending changes to `dev`, but the good news is they actually align fairly well I think!

mdr223 (commented):
Great job! I left mostly minor comments; the main work will be to merge my changes into this branch after Siva finishes his review of my branch -- but I actually think our work is fairly well-aligned! :)

demos/annotation.py
chjuncn: yeah, asked help from Gerardo to put them on the server .

src/palimpzest/core/data/datareaders.py
chjuncn: I can revert this variable name to "filename".

datareader is the place to convert what users' input to what the system needs, so if we don't keep the real user inputs in datareader, we lost this information in the whole system. And I think "data path" is important information, even more important than "file name". 

Do you mean we shouldn't keep the full path in datareader or you don't like the naming? 

For ValidationData, we can simply put this data path to execution_data logs, then everything will be fine.

src/palimpzest/core/data/validationdata.py
chjuncn: Yes, just trying to get the last item from plan_stats. I see you put sentinel_plan to another field, I think this still apply? What's the recommended way?

src/palimpzest/core/data/validationdata.py
chjuncn: done

PR #163, opened 2025-03-20
Merging in Changes for Sentinel Progress Bars; Split Convert (off by default); `demos/enron-demo.py`; and MMQA Benchmark

mdr223:
See title


src/palimpzest/core/lib/schemas.py
```diff
@@ -348,6 +348,10 @@ class File(Schema):
     filename = StringField(desc="The UNIX-style name of the file")
     contents = BytesField(desc="The contents of the file")
 
+class TextFile(Schema):
```
sivaprasadsudhir, line 351: Curious, why is this needed?

demos/cuad-demo.py
```diff
@@ -390,33 +396,31 @@ def compute_precision_recall(label_df, preds_df):
     return precision, recall
 
 
-# Score function for PZ optimizer.
-# Compare the predictions and labels for schema field.
-def score_fn(preds, labels):
-    assert isinstance(labels, list)
-    preds = handle_empty_preds(preds)
+# # Score function for PZ optimizer.
```
sivaprasadsudhir, line 399: delete this commented part?

demos/cuad-query-dep-demo.py
```diff
@@ -0,0 +1,665 @@
+import argparse
```
sivaprasadsudhir, line 1: what is this file for?

demos/enron-demo.py
```diff
@@ -0,0 +1,75 @@
...
+        item = {"fields": {}, "labels": {}}
+        item["fields"]["filename"] = filename
+        item["fields"]["contents"] = contents
+        item["labels"] = self.filename_to_labels.get(filename, {})
+
+        return item
+
+if __name__ == "__main__":
+    # create validation data source
+    val_datareader = EnronReader(dir="testdata/enron-eval-medium", labels_file="enron-eval-medium-labels.json", split="train")
```
sivaprasadsudhir, line 38: Should we add these files to testdata/ folder? If not, we should at least add them to the download script.

demos/enron-demo.py
```diff
@@ -0,0 +1,75 @@
...
+    plan = plan.sem_filter("The email is not quoting from a news article or an article written by someone outside of Enron")
+
+    # execute pz plan
+    config = pz.QueryProcessorConfig(
+        policy=pz.MaxQuality(),
+        val_datasource=val_datareader,
+        execution_strategy="parallel",
+        max_workers=20,
+    )
+    output = plan.run(config=config, k=5, j=3, sample_budget=50)
```
sivaprasadsudhir, line 57: Should we add k, j, and sample_budget to Config with default values?

src/palimpzest/query/operators/retrieve.py
```diff
@@ -200,29 +221,36 @@ def __call__(self, candidate: DataRecord) -> DataRecordSet:
         if query_is_str:
             query = [query]
 
-        # compute embedding(s) if the index is a chromadb collection
-        gen_stats = GenerationStats()
+        # compute input/query embedding(s) if the index is a chromadb collection
+        embeddings, gen_stats = None, GenerationStats()
```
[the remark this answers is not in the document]
mdr223: noted, I can lint those types of lines in the future

src/palimpzest/query/generators/generators.py
```diff
@@ -372,10 +374,13 @@ def __call__(self, candidate: DataRecord, fields: dict[str, Field] | None, json_
         # and can only account for the time spent performing the failed generation
         except Exception as e:
             logger.error(f"Error generating completion: {e}")
+            # print(f"Error generating completion: {e}")
```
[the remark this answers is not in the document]
mdr223: fixed this

demos/cuad-query-dep-demo.py
mdr223: This will ultimately hold a second version of the CUAD demo which will have filters based on supplemental annotations; it's not completed yet, hence the copy.

demos/enron-demo.py
mdr223: good call, I'll do that now

demos/enron-demo.py
mdr223: I'm on the fence about this and could go either way. On the one hand, we have a global config that just has everything in it, so what's the harm in adding more things. On the other hand, these are very specific to MAB sentinel execution.

src/palimpzest/core/lib/schemas.py
mdr223: The current `File` schema has a bytes field instead of a string field; some legacy demos break without the `TextFile` having a string field, and I think others require having bytes (i.e. they use `File`). So for now we just have a copy of each for backwards compatibility.

src/palimpzest/query/operators/retrieve.py
```diff
@@ -5,13 +5,16 @@
...
+from chromadb.utils.embedding_functions import SentenceTransformerEmbeddingFunction
 from chromadb.utils.embedding_functions.openai_embedding_function import OpenAIEmbeddingFunction
 from openai import OpenAI
 from ragatouille.RAGPretrainedModel import RAGPretrainedModel
+from sentence_transformers import SentenceTransformer
 
 from palimpzest.constants import MODEL_CARDS, Model
 from palimpzest.core.data.dataclasses import GenerationStats, OperatorCostEstimates, RecordOpStats
 from palimpzest.core.elements.records import DataRecord, DataRecordSet
+from palimpzest.core.lib.schemas import Schema
```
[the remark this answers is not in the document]
mdr223: I'll hold off on merging until I resolve this^, because I unless the issue is much bigger than what I'm seeing, I don't want to have to try cherry picking results to both (1) support Pneuma and (2) get this enron demo out to Mike's friend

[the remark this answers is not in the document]
sivaprasadsudhir: I think it should work. Basically, we should support also strings as input in the default case.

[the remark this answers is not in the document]
mdr223: Just pushed changes to this effect; will try running the pneuma demo this afternoon and if it works I'll merge.

PR #254, opened 2025-11-26
jczhang add model checks

xqlcn:


src/palimpzest/query/processor/query_processor_factory.py
```diff
@@ -91,6 +92,24 @@ def _config_validation_and_normalization(cls, config: QueryProcessorConfig, trai
         # set the final set of available models in the config
         config.available_models = available_models
```
mdr223: Just add a check here to ensure that `config.available_models` is not an empty list. If it is, raise a `ValueError`

mdr223 (changes requested):

PR #255, opened 2025-12-02
Semantic group by 253

mdr223:


src/palimpzest/query/operators/logical.py
```diff
@@ -381,33 +381,79 @@ def get_logical_op_params(self) -> dict:
...
+            self.agg_funcs = group_by_sig.agg_funcs
+        else:
+            # New API: using direct parameters
+            if gby_fields is None or agg_fields is None or agg_funcs is None:
+                raise ValueError("Must provide either group_by_sig or all of (gby_fields, agg_fields, agg_funcs)")
+            
+            # Validate fields exist in input schema
+            for f in gby_fields:
+                if f not in self.input_schema.model_fields:
+                    raise TypeError(f"Supplied schema has no field {f}")
```
mdr223: We can drop this check as semantic group by operations can group over fields which do not exist in the input schema. In our example where we are using a semantic group by to group complaints from customer reviews, the `complaint` field does not yet exist in the input schema. (The `contents` field does exist however, and we must have the `agg_fields` present in the input schema; hence the second check should remain).

src/palimpzest/core/elements/groupbysig.py
mdr223: It's okay to remove this file entirely!

tests/pytest/test_semantic_groupby.py
```diff
@@ -0,0 +1,127 @@
...
+]
+
+def test_semantic_groupby_basic():
+    """Test basic semantic group by functionality using the physical operator directly."""
+    print("Testing SemanticGroupByOp basic functionality...")
+    
+    try:
+        # Create list of candidates from text file dataset with schema
+        ds = pz.TextFileDataset(id="reviews", path="product-reviews/")
+        ds = ds.sem_map(review_cols)  # Add schema to extract complaint types
```
mdr223: You shouldn't need to use a `sem_map` to extract the `complaint` field. My guess is you needed this before because of the validation check for `gby_fields` in the `input_schema`, but if you remove that check then the test example I shared in the Github issue should work without the `sem_map`

tests/pytest/test_semantic_groupby.py
```diff
@@ -0,0 +1,127 @@
...
+        
+        print(f"Loaded {len(candidates)} review candidates with schema")
+        print(f"Sample candidate fields: {list(candidates[0].to_dict().keys()) if candidates else 'none'}")
+        
+        # Get input schema from the candidates
+        input_schema = candidates[0].schema if candidates else None
+        
+        # Create output schema (group by field + count)
+        from palimpzest.core.lib.schemas import create_schema_from_fields
+        output_schema = create_schema_from_fields([
```
mdr223: My apologies for not catching that you would need an `input_schema` and `output_schema` sooner, you should re-use this code from `datasetes.py` for your `output_schema` creation:
```
# Add group by fields to output schema
        for g in gby_fields:
            f = {"name": g, "type": Any, "desc": f"Group by field: {g}"}
            fields.append(f)

        # Add aggregation fields to output schema
        for i, agg_func in enumerate(agg_funcs):
            agg_field_name = f"{agg_func}({agg_fields[i]})"
            f = {"name": agg_field_name, "type": Any, "desc": f"Aggregate field: {agg_field_name}"}
            fields.append(f)

        output_schema = create_schema_from_fields(fields)
```

src/palimpzest/query/operators/aggregate.py
```diff
@@ -34,12 +34,37 @@ def __call__(self, candidates: list[DataRecord]) -> DataRecordSet:
...
-    and applies a function to each group. The group_by_sig object contains the fields to
-    group by and the aggregation functions to apply to each group.
+    and applies a function to each group.
+    
+    Can be initialized in two ways:
+    1. Legacy: group_by_sig parameter containing fields and functions
+    2. New: gby_fields, agg_fields, agg_funcs parameters directly
     """
-    def __init__(self, group_by_sig: GroupBySig, *args, **kwargs):
+    def __init__(self, group_by_sig: GroupBySig = None, gby_fields: list[str] = None,
```
mdr223: Similar to above; feel free to drop the `group_by_sig` argument entirely

src/palimpzest/query/operators/aggregate.py
```diff
@@ -664,3 +689,245 @@ def __call__(self, candidates: list[DataRecord]) -> DataRecordSet:
...
+    of fields and applies aggregation functions to each group using an LLM to determine the groups.
+    """
+    def __init__(self, gby_fields: list[str], agg_fields: list[str], agg_funcs: list[str], 
+                 model: Model | None = None, prompt_strategy: PromptStrategy = PromptStrategy.AGG, 
+                 reasoning_effort: str | None = None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.gby_fields = gby_fields
+        self.agg_fields = agg_fields
+        self.agg_funcs = agg_funcs
+        self.model = model if model is not None else Model.GPT_4O_MINI
```
mdr223: You can safely assume that model will never be `None`, so this line can become a simple assignment!

src/palimpzest/query/operators/aggregate.py
```diff
@@ -664,3 +689,245 @@ def __call__(self, candidates: list[DataRecord]) -> DataRecordSet:
...
+    
+    def get_model_name(self) -> str:
+        return self.model.value
+    
+    def get_fields_to_generate(self, candidate: DataRecord) -> list[str]:
+        """
+        For aggregation operators, we need to generate ALL output fields (including group-by fields),
+        not just the new fields. This overrides the default behavior.
+        """
+        return list(self.output_schema.model_fields.keys())
```
mdr223: One quirk about PZ's schemas is that they are additive. For example, if I do:
```
ds = pz.Dataset(...)
ds1 = ds.sem_map(cols=[{"name": "summary", ...}])
ds2 = ds1.sem_map(cols=[{"name": "title", ...}])
```
The `self.output_schema.model_fields` for the second `sem_map` would contain both `summary` and `title`; however this operation only generates `title`!

Taken together with my comment above re: keeping the `agg_field_name = agg_fields[i]` the base class implementation of `PhysicalOperator.get_fields_to_generate()` _should (in theory)_ suffice here. Thus, I think you can safely delete this function.

src/palimpzest/query/operators/aggregate.py
```diff
@@ -664,3 +689,245 @@ def __call__(self, candidates: list[DataRecord]) -> DataRecordSet:
...
+            }
+            
+            # Generate aggregation results
+            field_answers, _, agg_stats, _ = self.generator(group_members, fields, **gen_kwargs)
+            agg_results = {field: field_answers[field][0] for field in fields_to_generate}
+            
+            # Add aggregation results to data_item
+            for agg_field, agg_func in zip(self.agg_fields, self.agg_funcs):
+                output_field_name = agg_func.lower()
+                data_item[output_field_name] = agg_results.get(output_field_name)
```
mdr223: Long-term, we will want to support semantic aggregates which require using the generator to apply an LLM to compute the aggregate. In the short-term, let's just focus on a pre-defined set of non-semantic aggregates (e.g. `count`, `max`, `min`, `avg`, `list`, etc.) which are currently defined here: https://github.com/mitdbg/palimpzest/blob/2d8201e78803acca52d5b9cf442f65738ba3d279/src/palimpzest/query/operators/aggregate.py#L66-L123

This also means that (for now) you won't need to update the `gen_stats` beyond what is computed in the group creation phase.

mdr223 (commented):
Overall -- fantastic job!! You went above and beyond what I expected; including getting started with (1) using our `Generator` for LLM calls w/built-in cost / latency tracking and (2) marshaling those stats into a `DataRecordSet` to be returned by the `__call__` method.

That alone knocked out like 50-70% of what I had planned for your next steps. I'll follow-up with those details in the same issue from before.

src/palimpzest/core/data/dataset.py
```diff
@@ -573,10 +573,54 @@ def max(self) -> Dataset:
...
+        fields = []
+        
+        # Add group by fields to output schema
+        for g in gby_fields:
+            f = {"name": g, "type": Any, "desc": f"Group by field: {g}"}
+            fields.append(f)
+        
+        # Add aggregation fields to output schema
+        for i, agg_func in enumerate(agg_funcs):
+            agg_field_name = f"{agg_func}({agg_fields[i]})"
```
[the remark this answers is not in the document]
kepler11c: Hi Matt, 

This is a feature in both sem_groupby (which I created) and groupby (already existed in the codebase). Should I update both functions to simply leave the aggregation field name as is?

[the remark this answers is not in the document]
mdr223: Yes! Good catch, we should update both functions to simply use the `agg_field` for the `agg_field_name` 👍

src/palimpzest/core/data/dataset.py
```diff
@@ -572,10 +571,26 @@ def max(self) -> Dataset:
...
+            f = {"name": agg_field_name, "type": Any, "desc": f"Aggregate field: {agg_field_name}"}
+            fields.append(f)
+        
+        output_schema = create_schema_from_fields(fields)
```
mdr223: Since we duplicate this code in `.groupby()` and `.sem_groupby()` (below), I might recommend adding this as a function to `palimpzest.core.lib.schemas`:
```python
def create_groupby_schema_from_fields(gby_fields: list[str], agg_fields: list[str]):
    # construct the output schema dynamically based on groupby and aggregate fields
    fields = []
    
    # add group by fields to output schema
    for g in gby_fields:
        f = {"name": g, "type": Any, "desc": f"Group by field: {g}"}
        fields.append(f)
    
    # add aggregation fields to output schema
    for agg_field in agg_fields:
        f = {"name": agg_field, "type": Any, "desc": f"Aggregate field: {agg_field}"}
        fields.append(f)

    return create_schema_from_fields(fields)
```
Then you can simply import this function and use it here and below!

tests/pytest/test_semantic_groupby.py
```diff
@@ -34,11 +33,25 @@ def test_semantic_groupby_basic():
...
+        # Add group by fields to output schema
+        for g in ['complaint']:
+            f = {"name": g, "type": Any, "desc": f"Group by field: {g}"}
+            fields.append(f)
+        
+        # Add aggregation fields to output schema
+        agg_fields_list = ['contents']
+        agg_funcs_list = ['count']
+        for i, agg_func in enumerate(agg_funcs_list):
+            agg_field_name = f"{agg_func}({agg_fields_list[i]})"
```
mdr223: In keeping with changes elsewhere, we will want to update the `agg_field_name` to just be the `agg_fields_list[i]`

src/palimpzest/query/operators/aggregate.py
```diff
@@ -859,18 +851,18 @@ def __call__(self, candidates: list[DataRecord]) -> DataRecordSet:
                 full_op_id=self.get_full_op_id(),
                 logical_op_id=self.logical_op_id or "semantic-groupby",
                 op_name=self.op_name(),
-                time_per_record=(time.time() - start_time) / len(groups),
-                cost_per_record=total_cost / len(groups),
+                time_per_record=(time.time() - start_time) / len(agg_state),
+                cost_per_record=gen_stats.total_output_cost / len(agg_state),
```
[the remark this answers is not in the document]
kepler11c: Just pushed to changes to b1d8861a3be50e68b3ea4d7ce80cd84eb893dd05

Issue #274, opened 2026-01-17
`AssertionError: Output schema must be provided for convert prompts` when running `enron-demo.py`


khalefa-ow:
### Description

I am trying to run the Palimpzest demos, specifically `enron-demo.py`, but I consistently hit an `AssertionError` during execution.

The optimization phase completes successfully, but the program crashes at runtime with the error:

> `AssertionError: Output schema must be provided for convert prompts.`

---

### Environment

* **OS**: Ubuntu 22.04
* **Python versions tested**:

  * Python 3.12
  * Python 3.10
* **Palimpzest version**: 1.3.1
* **Execution method**: Virtual environment (`.venv`)

---

### Steps to Reproduce

1. Clone or download Palimpzest
2. Set up a virtual environment
3. Install dependencies
4. Run the demo:

```bash
python enron-demo.py
```

---

### Expected Behavior

The `enron-demo.py` script should complete successfully and produce output as described in the demo.

---

### Actual Behavior

The script fails with an `AssertionError` after optimization completes.

---

### Full Traceback

```
Total opt. time: 31.25s
Total opt. cost: $0.0207
Traceback (most recent call last):
  File "/home/mk/palimpzest_test_1.3.1/enron-demo.py", line 92, in <module>
    output = plan.optimize_and_run(train_dataset=train_dataset, validator=validator, config=config)
  File "/home/mk/palimpzest_test_1.3.1/.venv/lib/python3.12/site-packages/palimpzest/core/data/dataset.py", line 724, in optimize_and_run
    return QueryProcessorFactory.create_and_run_processor(self, config, train_dataset, validator)
  File "/home/mk/palimpzest_test_1.3.1/.venv/lib/python3.12/site-packages/palimpzest/query/processor/query_processor_factory.py", line 207, in create_and_run_processor
    return processor.execute()
  File "/home/mk/palimpzest_test_1.3.1/.venv/lib/python3.12/site-packages/palimpzest/query/processor/query_processor.py", line 124, in execute
    sentinel_plan_stats = self.sentinel_execution_strategy.execute_sentinel_plan(
        sentinel_plan, self.train_dataset, self.validator
    )
  File "/home/mk/palimpzest_test_1.3.1/.venv/lib/python3.12/site-packages/palimpzest/query/execution/mab_execution_strategy.py", line 838, in execute_sentinel_plan
    plan_stats = self._execute_sentinel_plan(plan, op_frontiers, validator, plan_stats)
[...]

PR #279, opened 2026-02-04
Fixing prompt strategy choice in RagRule and CritiqueAndRefine Rule 

SoTrx:
Hey,
First of all, thanks for your work!

While exploring the codebase, I noticed some errors when running a few examples that use optimization.
More specifically, some filter-like rules had a PromptStrategy.MAP attached to their generators, which then looked for an output_schema argument.

The cause seems to be the prompt_strategy = (...) line affecting PromptStrategy.MAP when use_reasoning_prompt is True, even if the physical operator is a Filter.

I believe this is also the cause of #274


mdr223:
Hi @SoTrx, thanks for tackling this issue! I somehow missed #274 in my notifications so I appreciate your timely help!

PR #301, opened 2026-04-13
Added policy parametrization from kwargs

vitaglianog:
Fixing a minor bug where when calling the `optimize_and_run()` method with a policy kwarg, e.g., `dataset.optimize_and_run(max_quality=True, validator=vdt, config=pz_config)`, the resulting code was parsing the policy but not adding it to the `config` object. 
Added the following line after the call to `policy = construct_policy_from_kwargs(**kwargs)`:

`config.policy = policy`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "palimpzest"
-version = "1.5.2"
+version = "1.5.3"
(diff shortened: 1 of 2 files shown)


mdr223 (approved):
Approved!

